PR #26, opened 2024-03-08
Catch hard print in Braket conversion

MoHermes:
braket-warning


Henri-ColibrITD (changes requested):
The output comes from their logger (it's raised in `braket\default_simulator\openqasm\interpreter.py`). We need to catch it (the logger warning), and redirect it as a standard warning

PR #29, opened 2024-03-11
Add identity gates on non-operated qubits 

hJaffaliColibritd:
When translating the circuit to a Braket circuit, to avoid non-contiguous qubit error at execution


Henri-ColibrITD (approved):
looks good and tests pass

PR #38, opened 2024-04-05
Functional tests execution in ideal case

hJaffaliColibritd:


mpqp/tools/maths.py
```diff
@@ -182,3 +185,69 @@ def exp(angle: Expr | Complex) -> sp.Expr | complex:
         res = sp.exp(angle)
         assert isinstance(res, Expr)
         return res
+
+
+def rand_orthogonal_matrix_seed(size: int, seed: int) -> Matrix:
```
Henri-ColibrITD: if all these are only used in one file, I'd move them to this file

Henri-ColibrITD: for the rest of the remarks, they are relatively unremarkable so I implemented them directly, see commits a59218c and 5f96d02

Henri-ColibrITD (changes requested):

PR #39, opened 2024-04-09
 inconsistent result formatting

JulienCalistoTD:


Henri-ColibrITD:
Should we add tests to stabilize this feature ?

PR #40, opened 2024-04-19
Allow exit with ctrl+c

MoHermes:
On connection setup, keyboard interrupt should be caught to gracefully quit .

### DIFF
--- a/docs/execution-extras.rst
+++ b/docs/execution-extras.rst
@@ -64,7 +64,7 @@ __________
 Connection setup script 
 -----------------------
 
-.. automodule:: mpqp.execution.connection.setup_connections
+.. automodule:: mpqp_scripts.setup_connections
 
 On disk configuration manager
 -----------------------------
--- a/mpqp/__init__.py
+++ b/mpqp/__init__.py
@@ -1,5 +1,5 @@
 # pyright: reportUnusedImport=false
 from mpqp.core.circuit import QCircuit
+from mpqp.core.instruction import Instruction
 from mpqp.core.instruction.barrier import Barrier
 from mpqp.core.languages import Language
-from mpqp.core.instruction import Instruction
--- a/mpqp/tools/choice_tree.py
+++ b/mpqp/tools/choice_tree.py
@@ -50,24 +50,28 @@ def run_choice_tree(question: QuestionNode):
     Args:
         question: Root question from which the choice tree will start
     """
-    # 3M-TODO: allow exit with Ctrl+C and Q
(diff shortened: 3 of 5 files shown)


hJaffaliColibritd (changes requested):
Tested the script while checking out this branch, but nothing happens when I press Ctrl+C while being in the main menu, Neither any other key combination. Tested on a windows console. Works fine when I select a choice of configuration (ibm, aws, ...)

image

hJaffaliColibritd:
image

On Linux console (Docker image with the current code), KeyboardInterrupt catched but raised apparently

henridb:
> Tested the script while checking out this branch, but nothing happens when I press Ctrl+C while being in the main menu, Neither any other key combination. Tested on a windows console. Works fine when I select a choice of configuration (ibm, aws, ...)

to put more precise words on the problem: we could call the two modes described the `pick` mode (when we use the library of the same name to pick a choice), and the `I/O` mode (when we are in the terminal display/input mode). So the problem you're describing is that the `pick` mode seems to capture the `KeyboardInterrupt` and stops it from having the usual effect (quitting the program).

Henri-ColibrITD:
> Tested the script while checking out this branch, but nothing happens when I press Ctrl+C while being in the main menu, Neither any other key combination. Tested on a windows console. Works fine when I select a choice of configuration (ibm, aws, ...)

to put more precise words on the problem: we could call the two modes described the `pick` mode (when we use the library of the same name to pick a choice), and the `I/O` mode (when we are in the terminal display/input mode). 

So the problem you're describing is that the `pick` mode seems to capture the `KeyboardInterrupt` and stops it from having the usual effect (quitting the program).

Henri-ColibrITD:
> Tested the script while checking out this branch, but nothing happens when I press Ctrl+C while being in the main menu, Neither any other key combination. Tested on a windows console. Works fine when I select a choice of configuration (ibm, aws, ...)

I confirm that this problem arises on my machine too

Henri-ColibrITD (changes requested):
feature not working

hJaffaliColibritd:
> Another problem @hJaffaliColibritd described (although for some reason I don't see it in this feed) is that the `I/O` mode doesn't exit gracefully on Linux (tested on a docker container)

I checked back on docker and Linux, it works, so I removed my comment in this PR

Henri-ColibrITD (approved):
looks good to me now

hJaffaliColibritd (approved):

PR #41, opened 2024-04-29
Dev

Henri-ColibrITD:
Greatly decreased import time (from 5s to 2s) by importing quantum SDKs as late as possible (i.e. only when a gate is needed for instance)


Henri-ColibrITD:
Wrong PR direction

PR #57, opened 2024-06-18
feat: remove target for Noise model and basis measure

JulienCalistoTD:


Henri-ColibrITD:
For now, we stop adding new features to dev until we push the fixes to main. I will review this PR then

Henri-ColibrITD (changes requested):

mpqp/core/instruction/measurement/basis_measure.py
```diff
@@ -49,12 +50,14 @@ class BasisMeasure(Measure):
 
     def __init__(
         self,
-        targets: list[int],
+        targets: Optional[list[int]] = None,
```
hJaffaliColibritd, line 53: We didn't do the same for ExpectationMeaasure right ?

mpqp/core/instruction/measurement/basis_measure.py
```diff
@@ -49,12 +50,14 @@ class BasisMeasure(Measure):
...
         c_targets: Optional[list[int]] = None,
         shots: int = 0,
```
hJaffaliColibritd, line 55: I have a suggestion following one of Julien's remark one day. Saying that if I want a StateVector, I can not put a measure in the circuit, or also do BasisMeasure() (with this new feature, which will put shots=0, and measure all the qubits), but then it is really bringing something? 

While if someone pulls out the BasisMeasure, probably it is to specify the targets and still put shots=0, or specify the shots (and possibly the target).

In that sense, I propose that we put the shots default equal to 1024 or something like this, which will permit quckily to have a SAMPLE job for the whole circuit by just putting BasisMeasure().

This will be a bit a breaking change.

mpqp/noise/noise_model.py
```diff
@@ -142,6 +145,7 @@ class Depolarizing(NoiseModel):
              └───┘
         NoiseModel:
             Depolarizing(0.32, [0, 1, 2], 1)
+            Depolarizing(0.01, [all], 1)
```
hJaffaliColibritd, line 149: Why `all` in the array ?

hJaffaliColibritd (changes requested):

mpqp/core/instruction/measurement/basis_measure.py
JulienCalistoTD: Yes, only Basismeasure is allowed to have no target

mpqp/core/circuit.py
```diff
@@ -220,27 +220,107 @@ def add(self, components: OneOrMany[Instruction | NoiseModel]):
...
+                    isinstance(noise, Depolarizing)
+                    and len(noise.targets) < noise.dimension
+                ):
+                    raise ValueError(
+                        f"Number of target qubits {len(noise.targets)} should be higher than the dimension {noise.dimension}."
+                    )
+        qcircuit.noises = noises
+
+        instructions = deepcopy(self.instructions)
+        for instruction in instructions:
```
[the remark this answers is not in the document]
JulienCalistoTD: Only Basismeasure is allowed to have no target

mpqp/noise/noise_model.py
JulienCalistoTD: there are several solutions for print:
 - execute hard_copy to obtain the real target at step
 - print 'all' or 'dynamic', to show that it's for all qubits, so don't parse all instructions

mpqp/core/circuit.py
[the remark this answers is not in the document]
hJaffaliColibritd: Ok then I create a task for the ExpectationMeasure on Trello

mpqp/core/instruction/measurement/basis_measure.py
```diff
@@ -49,12 +50,14 @@ class BasisMeasure(Measure):
 
     def __init__(
         self,
-        targets: list[int],
+        targets: Optional[list[int]] = None,
         c_targets: Optional[list[int]] = None,
-        shots: int = 0,
+        shots: int = 1024,
```
hJaffaliColibritd, line 55: I think should discuss it before really doing it,it was just a proposition. @Henri-ColibrITD  @MoHermes look at the comment line 53

hJaffaliColibritd (changes requested):

mpqp/core/circuit.py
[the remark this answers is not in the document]
Henri-ColibrITD: but is `Measure` is already considered, why d we need to consider special cases ?

Henri-ColibrITD (approved):

PR #59, opened 2024-06-24
Statistical noisy tests

MoHermes:


hJaffaliColibritd (changes requested):
Didn't finish the review but put already few comments

mpqp/tools/theoretical_simulation.py
```diff
@@ -20,7 +20,7 @@ def process_qcircuit(circuit: QCircuit):
     gates = [
         (
             (gate.label, gate.controls + gate.targets)
-            if gate.label == "CNOT"
+            if isinstance(gate, ControlledGate)
```
MoHermes, line 23: @hJaffaliColibritd Modified for handling all the other controlled gates.

Henri-ColibrITD:
This is not ready for merge at all, it is a preliminary work, it will need one more sprint to be finalized, most likely

cbpt-rueil:
the noise is depolarizing and not dephasing, it would be nice to formally prove the alpha formula

Henri-ColibrITD:
the noise is depolarizing and not dephasing, it would be nice to formally prove the alpha formula

hJaffaliColibritd:
- Test for other noise models (Bitflip, ....)
- Generalize so device are in parameter of the tests

hJaffaliColibritd (changes requested):
Review part 1

hJaffaliColibritd:
Do we only handle noisy simulations for shot noise and probability distribution ? 
What about amplitude distribution (meaning error in the complex amplitudes) ? 

What about noise for expectation value ? At least shot noise ?

tests/tools/test_theoretical_simulation.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+
+from mpqp import QCircuit
+from mpqp.core.instruction.measurement import BasisMeasure
+from mpqp.gates import *
+from mpqp.tools.theoretical_simulation import theoretical_probs
+
+
+def test_simulation():
```
hJaffaliColibritd, line 9: What about testing the theoretical simulation in noisy case without comparing to the output of an SDK, but compared to a density matrix we computed analytically (by hand or using Maple for instance ?)

tests/noise/test_noisy_execution.py
```diff
@@ -12,10 +14,27 @@
...
+from mpqp.tools.theoretical_simulation import validate_noisy_circuit
+
+noisy_devices: list[Any] = [
+    dev
+    for dev in list(ATOSDevice) + list(AWSDevice) + list(IBMDevice) + list(GOOGLEDevice)
+    if dev.is_noisy_simulator()
+]
+# TODO: in the end this should be automatic as drafted above, but for now only
+# one device is stable
+noisy_devices = [AWSDevice.BRAKET_LOCAL_SIMULATOR]
```
hJaffaliColibritd, line 37: We can add IBMDevice.AER_SIMULATOR now for instance

hJaffaliColibritd (changes requested):

mpqp/noise/noise_model.py
```diff
@@ -1,24 +1,31 @@
...
-import numpy as np
 from typeguard import typechecked
 
 from mpqp.core.instruction.gates.native_gates import NativeGate
 from mpqp.core.languages import Language
-from mpqp.noise.custom_noise import KrausRepresentation
 from mpqp.tools.generics import T
 
 
 def plural_marker(items: Sequence[T]):
```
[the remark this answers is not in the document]
Henri-ColibrITD: I've added documantation but I think this is a pretty niche function, so I've
set it to be a private function instead of putting it in tools. Tell me if you
agree with this choice.

mpqp/noise/noise_model.py
```diff
@@ -114,9 +163,16 @@ def to_other_language(
         """
         pass
 
-    def info(self, qubits: set[int]) -> str:
+    def pprint(self) -> str:
```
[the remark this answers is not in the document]
Henri-ColibrITD: good point

mpqp/noise/noise_model.py
```diff
@@ -93,10 +100,52 @@ def connections(self) -> set[int]:
...
+        of your circuit, and the targets involved. In particular, the targets of
+        the noise application may not match the noise targets, because the noise
+        targets signifies all the qubits that the noise is applicable on, but if
+        the noise happens at a gate execution, it would only actually impact the
+        targets qubits of the gate.
+
+        Note:
+            This generic method considers that the default Kraus operators of
+            the noise are for one qubit noises. If this is not the case, this
+            method should be overloaded in the corresponding class.
```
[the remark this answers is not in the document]
Henri-ColibrITD: hmmm, to make sure I understand, let me take an example:
for a non dimensional noise, the operators would be p1II, p2IK, p3KI and p4KK,
while for a dimensional it would just be p1II and p2K2

mpqp/tools/theoretical_simulation.py
```diff
@@ -0,0 +1,210 @@
...
+        CZ(1, 2),
+    ]
+
+    probs = [0.001, 0.01, 0.1, 0.1, 0.2, 0.3]
+    xs = sum(([elt] * 6 for elt in probs), start=[])
+    shots_vals = [500, 1_000, 5_000, 10_000, 50_000, 100_000]
+    ys = shots_vals * 6
+    dists = np.array(
+        [
+            exp_id_dist_excess(QCircuit(gates + [Depolarizing(prob)]), shots)
```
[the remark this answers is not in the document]
Henri-ColibrITD: not sure we need to overthink this: it's a manual test to debug my
implementation, I could even remove it if you want since this should be formally
tested

tests/noise/test_noisy_execution.py
Henri-ColibrITD: done

tests/noise/test_noisy_execution.py
```diff
@@ -96,16 +119,34 @@ def test_all_native_gates_global_noise_execution_without_error(
...
+    product(
+        [0.001, 0.01, 0.1, 0.1, 0.2, 0.3],
+        [500, 1_000, 5_000, 10_000, 50_000, 100_000],
+        noisy_devices,
+    ),
+)
+def test_validate_noise(
+    circuit: QCircuit, depol_noise: float, shots: int, device: AvailableDevice
+):
+    circuit.add(Depolarizing(depol_noise))
```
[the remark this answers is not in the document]
Henri-ColibrITD: God point. I'll do that once the Kraus operator problem for dimensional noises
problem is sorted out

mpqp/noise/noise_model.py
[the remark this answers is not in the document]
hJaffaliColibritd: Ok to put it private

mpqp/tools/theoretical_simulation.py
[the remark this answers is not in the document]
hJaffaliColibritd: I think we can let it like this yeah. I mark as resolved

tests/tools/test_theoretical_simulation.py
Henri-ColibrITD: this will come in a future PR

mpqp/noise/noise_model.py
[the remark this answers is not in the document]
Henri-ColibrITD: since this is a bit complex it will come in a future PR so that we can merge this one

mpqp/tools/theoretical_simulation.py
```diff
@@ -0,0 +1,210 @@
...
+                    (
+                        k @ state @ k.T.conj()
+                        for k in noise.to_adjusted_kraus_operators(
+                            gate.connections(), circ.nb_qubits
+                        )
+                    ),
+                    start=np.zeros((d, d), dtype=np.complex64),
+                )
+
+    connected_qubits = set().union(*[gate.connections() for gate in circ.get_gates()])
```
hJaffaliColibritd, line 54: Since we call two times `circ.get_gates()` here and in line 35, it can be interesting to call it only once, and store it ?

mpqp/tools/theoretical_simulation.py
```diff
@@ -0,0 +1,210 @@
...
+    unconnected_qubits = set(range(circ.nb_qubits)).difference(connected_qubits)
+    for noise in circ.noises:
+        if len(noise.gates) == 0:
+            sum(
```
hJaffaliColibritd, line 58: the sum is not stored in a variable ?

mpqp/tools/theoretical_simulation.py
```diff
@@ -0,0 +1,210 @@
...
+    without noise and the noisy circuits probability distributions. This
+    interval depends on this distance in a non linear manner. This function
+    gives the relation between the two.
+
+    Args:
+        alpha: The distance Jensen-Shannon between the non noisy distribution
+            and the noisy distribution.
+
+    Returns:
+        The trust interval for the distance.
```
hJaffaliColibritd, line 82: We don't really return here an interval, but rather the size of the trust interval, right?

mpqp/tools/theoretical_simulation.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    return np.sqrt(alpha)
+
+
+# TODO: this bound it coming out of my ass, see if these results already in
+# statistics. Some useful resources might be: jensen inequality, chernoff bound
+
+
+def trust_int(circuit: QCircuit):
+    """Given a circuit, this computes the trust interval for the output samples
```
hJaffaliColibritd, line 92: Does it compute the interval, or the size of the interval (see line 99)?

mpqp/tools/theoretical_simulation.py
```diff
@@ -0,0 +1,210 @@
...
+    given into consideration the noise in the circuit.
+
+    Args:
+        circuit: The circuit.
+
+    Returns:
+        The size of the trust interval (related to the Jensen-Shannon distance).
+    """
+    noiseless_circuit = QCircuit(
```
hJaffaliColibritd, line 101: There is no `NoiseModel` object in `circuit.instructions`, because noise models are stored in `circuit.noises`.

I would recommand the usage of `circuit.without_noises()` instead

mpqp/tools/theoretical_simulation.py
```diff
@@ -0,0 +1,210 @@
...
+        circuit: The circuit (with potential noises).
+        shots: Number of shots in the basis measurement.
+
+    Returns:
+        The gap between:
+
+        1. the distance between theory and our results;
+        2. and the trust interval.
+    """
+    return exp_id_dist(circuit, shots) - trust_int(circuit)
```
hJaffaliColibritd, line 166: Should we return the absolute value of this difference?

hJaffaliColibritd (changes requested):

mpqp/tools/theoretical_simulation.py
Henri-ColibrITD: good point

mpqp/tools/theoretical_simulation.py
Henri-ColibrITD: oops

mpqp/tools/theoretical_simulation.py
Henri-ColibrITD: nitpicking much ? :p

mpqp/tools/theoretical_simulation.py
Henri-ColibrITD: good point

mpqp/tools/theoretical_simulation.py
Henri-ColibrITD: indeed, but overall I think we'll have to redo a pass on this trust interval
thing, at least to prove it formally

mpqp/tools/theoretical_simulation.py
hJaffaliColibritd: Not really ^^, I struggled a bit to catch the differences between the intervals, the several distances, the interval size etc. Maybe we can detail that in the onlline documentation, to put some theoretical context?

Henri-ColibrITD: good point, this should go in the header of this file I guess

Henri-ColibrITD: I added a big ol piece of documentation, I hope it will please you (for my standards, I was exceptionally verbose! You were right, it was needed)

hJaffaliColibritd (approved):

PR #62, opened 2024-06-26
New minor version, with it's flow of buxfixes

Henri-ColibrITD:


MoHermes (approved):

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -1,19 +1,28 @@
...
 module contains a few abstract classes used to factorize the behaviors common to
-a lot of gates. Feel free to use them for your own custom gates!"""
+a lot of gates.
+
+You will find bellow the list of available native gates:
+
+.. container::
+    :name: native-gates-list
+    
+    ``to-be-generated``
```
hJaffaliColibritd, line 11: Should we remove this ?

mpqp/noise/noise_model.py
```diff
@@ -30,14 +31,14 @@ class NoiseModel(ABC):
...
-        if len(targets) == 0:
-            raise ValueError("Expected non-empty target list")
-
+    def __init__(
+        self, targets: Optional[list[int]] = None, gates: Optional[list[type[Gate]]] = None
+    ):
+        if targets is None:
+            targets = []
         if len(set(targets)) != len(targets):
             raise ValueError(f"Duplicate indices in targets: {targets}")
```
hJaffaliColibritd, line 43: If we allow ``targets`` to be empty in the case when we want all qubits to be targets, then we will always enter this condition of duplicate indices, no? One workaround would be to check if len(targets)!=0 first

mpqp/noise/noise_model.py
```diff
@@ -178,7 +187,7 @@ def __init__(
                 )
 
         nb_targets = len(targets)
-        if nb_targets < dimension:
+        if nb_targets != 0 and nb_targets < dimension:
```
hJaffaliColibritd, line 190: The problem with this is that if targets==[], meaning we need all qubits of the circuit, we don't know how many exact qubit we will target at this point, so we cannot catch really an error with dimension.
Example, you have a 2-qubit circuit, you don't precise the target, so at runtime it will become [0,1], BUT, you have instantiated a noise model with dimension=3, here the problem will not be catched, meaning we need to do a double check somewhere else, or move this verification later in the execution ... ?

This requires for me some discussion

mpqp/tools/generics.py
```diff
@@ -205,4 +223,49 @@ class SimpleClassReprABCMeta(SimpleClassReprMeta, ABCMeta):
 
 
 class SimpleClassReprABC(metaclass=SimpleClassReprABCMeta):
-    pass
\ No newline at end of file
+    pass
```
hJaffaliColibritd, line 226: Should we put a Doc for this class, cauz I don't understand directly what it is useful for from the name

mpqp/tools/generics.py
```diff
@@ -205,4 +223,49 @@ class SimpleClassReprABCMeta(SimpleClassReprMeta, ABCMeta):
...
+
+
+class classproperty:
```
hJaffaliColibritd, line 229: Doc missing

requirements-dev.txt
```diff
@@ -1,6 +1,7 @@
 -r requirements.txt
 setuptools>=56.0.0
 pytest==7.1.1
+nbmake
```
hJaffaliColibritd, line 4: no version ?

hJaffaliColibritd (changes requested):

mpqp/noise/noise_model.py
hJaffaliColibritd: Indeed, my fault

requirements-dev.txt
Henri-ColibrITD: it's not needed so no

mpqp/noise/noise_model.py
JulienCalistoTD: It is solve in the branche remove target. But it will not comme in this release

requirements-dev.txt
JulienCalistoTD: it is a package to transforme notebook to executable

mpqp/noise/noise_model.py
hJaffaliColibritd: So in my sense we should remove the possibility to put target as optional and empty, and only include the whole working code when it will be the time for the right release

JulienCalistoTD: yes agree

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: no, it is a target container for the dynamically generated gate list. the list is generated in js, so this stays like this until someone loads in in there browser. At that point, the `to-be-generated` is replaced by said list

mpqp/noise/noise_model.py
Henri-ColibrITD: agreed

mpqp/tools/generics.py
Henri-ColibrITD: done

mpqp/core/circuit.py
```diff
@@ -223,13 +222,21 @@ def add(self, components: OneOrMany[Instruction | NoiseModel]):
...
             ]
             if basisMs and all([len(bm.targets) != self.nb_qubits for bm in basisMs]):
                 raise ValueError(
                     "In noisy circuits, BasisMeasure must span all qubits in the circuit."
                 )
+            
+            if isinstance(components, Depolarizing) and len(components.targets) < components.dimension:
+                 raise ValueError(
+                f"Number of target qubits {len(components.targets)} should be higher than the dimension {components.dimension}."
+            )
```
JulienCalistoTD, line 239: solve  nb_target < dimension

mpqp/noise/noise_model.py
Henri-ColibrITD: I like the solution proposed in 664b60d

JulienCalistoTD (approved):

mpqp/noise/noise_model.py
Henri-ColibrITD: after more thought, this solution wasn't complete enough and thus was reverted in 971d97c

hJaffaliColibritd (approved):

PR #68, opened 2024-07-01
Feat breakpoints

Henri-ColibrITD:


mpqp/core/circuit.py
```diff
@@ -222,6 +224,9 @@ def add(self, components: OneOrMany[Instruction | NoiseModel]):
             components.size = self.nb_qubits
 
         if isinstance(components, NoiseModel):
+            if len(components.targets) == 0:
+                components.targets = [target for target in range(self.nb_qubits)]
```
hJaffaliColibritd, line 228: This comes from a previous commit on dev, we said that the fact that target is empty will integrated by merging the other branch related to this feature. So this part should be removed

mpqp/core/instruction/breakpoint.py
```diff
@@ -0,0 +1,32 @@
...
+        enabled: A breakpoint can be kept in the circuit but temporarily
+            disabled using this argument.
+        label: A breakpoint can be given a label for easier traceability.
+    """
+
+    def __init__(
+        self,
+        draw_circuit: bool = False,
+        enabled: bool = True,
+        label: str | None = None,
```
hJaffaliColibritd, line 22: Do you think it is useful to keep the info about where this breakpoint is happening in the circuit (like position in the instruction array etc). Because for instance if you use `circuit.breakpoints()` how do you know which one is where exactly ? (you only have how they are ordered I think)

mpqp/core/circuit.py
Henri-ColibrITD: this code actually doesn't come from me. It's only marked new because I merged the branch it comes from with mine to check compatibility

mpqp/core/instruction/breakpoint.py
Henri-ColibrITD: indeed, this was a bit annoying to handle: in order to know the position, I had to iterate over the circuit's instructions to get the index. On the other hand, not sure I'd like to store the index of the bp. For instance, if I did
```py
c = QCircuit([Y(0), Breakpoint()])
del c.instructions[0]
```
then I would have the wrong stored index for the breakpoint. The alternative would be to recompute it each tome we modify the instructions, but I don't this this is a better solution

hJaffaliColibritd: Maybe a solution would be to save the reference of the previous gate ? In that case you can quickly re-find where you are putting a breakpoint in the circuti ?

hJaffaliColibritd (approved):

mpqp/core/circuit.py
Henri-ColibrITD: but it was a merge error, I'll remove it, thanks for the reminder :)

mpqp/core/instruction/breakpoint.py
Henri-ColibrITD: After a long discussion, we decided to carry on with the previous implementation, and only change it latter, if we realize it causes problems we couldn't foresee

PR #69, opened 2024-07-02
Fix cirq obs variance

JulienCalistoTD:


JulienCalistoTD:
write some test

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -309,6 +313,98 @@ def from_matrix(cls, matrix: Matrix) -> PauliString:
...
+        """
+        from cirq.ops.pauli_string import PauliString as CirqPauliString
+        from cirq.ops.linear_combinations import PauliSum as CirqPauliSum
+        
+        if isinstance(pauli, list) and all(isinstance(item, CirqPauliString) for item in pauli):
+            if nb_qubits == 0:
+                for pauli_mono in pauli:
+                    nb_qubits = max(PauliString._get_nb_qubits_cirq_pauli(pauli_mono), nb_qubits)
+            return [PauliString._from_cirq(pauli_mono, nb_qubits) for pauli_mono in pauli]
+        if isinstance(pauli, (CirqPauliSum, CirqPauliString)):
```
Henri-ColibrITD, line 402: is the max useful ? since you already checked that `nb_qubits == 0`

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -309,6 +313,98 @@ def from_matrix(cls, matrix: Matrix) -> PauliString:
...
+                nb_qubit = int(qubit.name.split("_")[1])
+                atoms[nb_qubit] = get_atom(op)
+            pauli_string += PauliStringMonomial(coef, atoms)
+
+        return pauli_string
+
+    @classmethod
+    def from_other_languages(
+        cls, pauli: Union[CirqPauliSum, CirqPauliString, List[CirqPauliString]], nb_qubits: int = 0
+    ) -> PauliString | List[PauliString]:
```
Henri-ColibrITD, line 380: why do we need to pass a list of PS ?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -309,6 +313,98 @@ def from_matrix(cls, matrix: Matrix) -> PauliString:
...
+        """
+        Convert pauli objects from other quantum computing libraries to PauliString.
+
+        args:
+            pauli: The pauli object(s) to be converted. This can be a single CirqPauliSum, a single CirqPauliString, or a list of CirqPauliString objects.
```
Henri-ColibrITD, line 385: no need to list the input types on the doc

mpqp/core/instruction/measurement/pauli_string.py
Henri-ColibrITD: we could also add PS from other SDKs but put them behind an `NotImplementedError`

Henri-ColibrITD (changes requested):
some questions to be answered

Henri-ColibrITD (approved):

PR #70, opened 2024-07-03
Feat qcirc to matrix and gate to matrix

JulienCalistoTD:


mpqp/core/instruction/gates/native_gates.py
```diff
@@ -150,7 +150,7 @@ def to_other_language(
     ):
         if qiskit_parameters is None:
             qiskit_parameters = set()
-        theta = float(self.theta) if self._numeric_parameters else self.theta
+        theta = float(self.theta)
```
JulienCalistoTD, line 153: theta can be float | Expr.
When we have Expr, it raises an error when converts to a float when we use to_matrix in Qiskit

JulienCalistoTD: need qiskit 1.0

JulienCalistoTD:
documentation

mpqp/core/instruction/gates/controlled_gate.py
```diff
@@ -39,3 +41,118 @@ def __init__(
         """See parameter description."""
 
         Gate.__init__(self, targets, label)
+
+    def controlled_gate_to_matrix(self) -> Matrix:
```
Henri-ColibrITD, line 45: why is this a separate method ? `to_matrix` wouldn't work ?

mpqp/core/instruction/gates/controlled_gate.py
```diff
@@ -39,3 +41,118 @@ def __init__(
...
+
+        for i in range(1, nb_qubit):
+            if i == control:
+                control_matrix = np.kron(control_matrix, zero)
+            else:
+                control_matrix = np.kron(control_matrix, I2)
+
+        return control_matrix + target_matrix
+
+    def _multi_control_gate_to_matrix(self) -> Matrix:
```
Henri-ColibrITD, line 91: it sounds like the previous method is a special case of this one

mpqp/core/instruction/gates/controlled_gate.py
```diff
@@ -39,3 +41,120 @@ def __init__(
...
+
+        I2 = np.eye(2)
+
+        min_qubit = min(min(self.controls), min(self.targets))
+        max_qubit = max(max(self.controls), max(self.targets))
+        nb_qubits = max_qubit - min_qubit + 1
+
+        canonical_matrix = self.to_canonical_matrix()
+        while canonical_matrix.shape[0] < 2**nb_qubits:
+            canonical_matrix = np.kron(canonical_matrix, I2)
```
Henri-ColibrITD, line 106: you'd have to kron more than once

mpqp/core/instruction/gates/controlled_gate.py
```diff
@@ -39,3 +41,120 @@ def __init__(
...
+
+        matrix = np.eye(2**nb_qubits, dtype=np.complex64)
+        qubit_types = {i: "None" for i in range(nb_qubits)}
+
+        for i, _ in enumerate(self.controls):
```
Henri-ColibrITD, line 111: why using enumerate here ?

Henri-ColibrITD (changes requested):
overall, some methods seem a bit over-complicated. I'll see if I can simplify it. We could also discuss this live because I may have missed important complexity points

mpqp/core/circuit.py
```diff
@@ -265,7 +265,7 @@ def hard_copy(self):
             c: 4/══════╩══╩═════╩══╩═
                        0  1     2  3
             NoiseModel:
-                Depolarizing(0.01, [all], 1)
+                Depolarizing(0.01)
             >>> circuit2.nb_qubits = 3
```
Henri-ColibrITD, line 269: I just noticed that the example given here is an example on how some instruction have a dynamic size, and is not at all about the method (it is not even called in the example!), this should be fixed

mpqp/core/instruction/gates/controlled_gate.py
```diff
@@ -39,3 +41,131 @@ def __init__(
...
+                for _ in range(1, start - 1):
+                    extended_swap_matrix = np.kron(extended_swap_matrix, I2)
+                extended_swap_matrix = np.kron(extended_swap_matrix, swap_matrix)
+            for _ in range(int(math.log2(extended_swap_matrix.shape[0])), nb_qubits):
+                extended_swap_matrix = np.kron(extended_swap_matrix, I2)
+            return np.dot(matrix, extended_swap_matrix)
+
+        for control in sorted(self.controls):
+            control -= min_qubit
+            if qubit_types[control] != "control":
```
Henri-ColibrITD, line 133: you're iterating over the controls, why wouldn't it be a control

Henri-ColibrITD:
Ok, after having one more look at it, I think I won't have to time to finish this before my leave, the controlled_gate to_matrix method for sure needs some love, I can take care of it once I'm back

mpqp/core/instruction/gates/controlled_gate.py
JulienCalistoTD: done

mpqp/core/instruction/gates/controlled_gate.py
JulienCalistoTD: yes, when we have multiple control or target we use a different methode to compute. start from canonical matrix and do swap

mpqp/core/instruction/gates/controlled_gate.py
JulienCalistoTD: done

mpqp/core/instruction/gates/controlled_gate.py
JulienCalistoTD: done

mpqp/core/instruction/gates/controlled_gate.py
```diff
@@ -39,3 +41,120 @@ def __init__(
...
+        min_qubit = min(min(self.controls), min(self.targets))
+        max_qubit = max(max(self.controls), max(self.targets))
+        nb_qubits = max_qubit - min_qubit + 1
+
+        canonical_matrix = self.to_canonical_matrix()
+        while canonical_matrix.shape[0] < 2**nb_qubits:
+            canonical_matrix = np.kron(canonical_matrix, I2)
+
+        matrix = np.eye(2**nb_qubits, dtype=np.complex64)
+        qubit_types = {i: "None" for i in range(nb_qubits)}
```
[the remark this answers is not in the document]
JulienCalistoTD: yes, to follow the targets and controls with swap

mpqp/core/instruction/gates/controlled_gate.py
JulienCalistoTD: its the case on multiple control

mpqp/core/circuit.py
JulienCalistoTD: For repr it is '[]', and for now, there is no str.

mpqp/core/instruction/gates/controlled_gate.py
```diff
@@ -39,3 +42,116 @@ def __init__(
...
+        swaps = []
+        # qubit_types if a representation of the all qubits to follow target and control with swap
+        # assuming that canonical_matrix start with control and then target
+        qubit_types = {i: "None" for i in range(nb_qubits)}
+        qubit_types.update({i: "control" for i in range(len(controls))})
+        qubit_types.update({i + len(controls): "target" for i in range(len(targets))})
+
+        def swap_and_update(qubits: list[int], target_type: str):
+            for qubit in sorted(qubits):
+                if qubit_types[qubit] != target_type:
```
Henri-ColibrITD, line 130: actually, this works for controls but not for targets, so this shoud be
reworked, I'll try to see that tomorrow

Henri-ColibrITD:
A last point is under examination: should we have a special case for "simple" control gates. And are all the needed test present ? if so, I'm good for merge

Henri-ColibrITD (approved):

mpqp/core/instruction/gates/gate_definition.py
```diff
@@ -43,6 +43,11 @@ class GateDefinition(ABC):
 
     @abstractmethod
     def to_matrix(self) -> Matrix:
+        """Returns the matrix corresponding to this gate definition. With
+        considering potential column and row permutations needed if the targets."""
```
hJaffaliColibritd, line 47: Sentence uncomplete

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -252,6 +255,9 @@ def to_other_language(
     def to_matrix(self) -> Matrix:
         return self.matrix
 
+    def to_canonical_matrix(self):
```
hJaffaliColibritd, line 258: return type missing, and why not putting self.matrix directly ?

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -479,6 +485,9 @@ def to_matrix(self) -> Matrix:
             ]
         )
 
+    def to_canonical_matrix(self):
```
hJaffaliColibritd, line 488: Return type missing ? (not mandatory, but we put it for to_matrix, so to stay coherent)

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -598,6 +608,51 @@ def __repr__(self) -> str:
         2
     )
 
+    def to_matrix(self, nb_qubits: int = 0) -> npt.NDArray[np.complex64]:
+        """Constructs the matrix representation of a SWAP gate for two qubits in
+        a multi-qubit system.
```
hJaffaliColibritd, line 613: multi-qubit QCircuit ?

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -598,6 +608,51 @@ def __repr__(self) -> str:
...
+
+        Args:
+            nb_qubits: The total number of qubits in the system. If not
```
hJaffaliColibritd, line 616: in the circuit ?

mpqp/noise/noise_model.py
```diff
@@ -163,7 +163,7 @@ class Depolarizing(NoiseModel):
              └───┘
         NoiseModel:
             Depolarizing(0.32, [0, 1, 2])
-            Depolarizing(0.01)
+            Depolarizing(0.01, [])
```
hJaffaliColibritd, line 166: maybe print the hardcopy of the circuit to avoid having an empty list ?

hJaffaliColibritd (changes requested):
First part of review

mpqp/noise/noise_model.py
```diff
@@ -454,7 +452,7 @@ def __init__(
     def to_kraus_representation(self) -> KrausRepresentation: ...
 
     def __repr__(self):
-        targets = f", targets={self.targets}" if len(self.targets) != 0 else ""
+        targets = f", targets={self.targets}"
         gates = f", gates={self.gates}" if len(self.gates) != 0 else ""
         prob = f", prob={self.prob}" if self.prob != 1 else ""
```
hJaffaliColibritd, line 457: if prob is not equal to 1, then empty string will be put after self.gamma in the __repr__, so we will display directly the third argument (targets), while we don't use keywords. I think this will result in a wrong repr.

Example : 
If I instantiate the object `a = AmplitudeDamping(0.3, targets=[0,1])`, the output repr will be `AmplitudeDamping(0.3, [0,1])`, which is a wrong __repr__.

mpqp/tools/circuit.py
```diff
@@ -0,0 +1,89 @@
...
+
+    Returns:
+        A quantum circuit with the specified number of qubits and randomly chosen gates.
+
+    Raises:
+        ValueError: If the number of qubits is too low for the specified gates.
+
+    Examples:
+        >>> random_circuit([U, TOF], 3) # doctest: +SKIP
+        Generates a random quantum circuit with 3 qubits using U and TOF gates.
```
hJaffaliColibritd, line 34: Why do we output a description of what it generates ? Can't we print the circuit, and still skip the doctest? (I understand it is hard to doctest random circuit generation)

mpqp/tools/generics.py
```diff
@@ -153,6 +154,51 @@ def find_index(iterable: Iterable[T], oracle: Callable[[T], bool]) -> int:
     raise ValueError("No objects satisfies the given oracle")
 
 
+def compute_expected_matrix(qcircuit: QCircuit):
```
hJaffaliColibritd, line 157: Shouldn't we put the input type `"QCircuit"` as a string here? (since we only import it in the if TYPE_CHECKING)

mpqp/tools/maths.py
```diff
@@ -62,15 +62,15 @@ def matrix_eq(lhs: Matrix, rhs: Matrix, atol: float = atol, rtol: float = rtol)
...
 
     for elt in zip(np.ndarray.flatten(lhs), np.ndarray.flatten(rhs)):
-        if isinstance(elt[0], Expr) or isinstance(elt[1], Expr):
-            if elt[0] != elt[1]:
-                return False
-        else:
+        try:
             if abs(elt[0] - elt[1]) > (atol + rtol * abs(elt[1])):
                 return False
+        except TypeError:
```
hJaffaliColibritd, line 70: Is this a clean way to handle the case when elements are of type `Expr`? What if we have another third case where the type is different from `Expr`, are we sure that comparing with the if is always working ?

tests/core/test_circuit.py
```diff
@@ -350,3 +353,42 @@ def test_to_qasm_3(circuit: QCircuit, printed_result_filename: str):
...
+            QCircuit([Rz(np.pi / 4, 0)]),
+            Rz(np.pi / 4, 0).to_matrix(),
+        ),
+        (
+            QCircuit([Ry(np.pi / 2, 0)]),
+            Ry(np.pi / 2, 0).to_matrix(),
+        ),
+    ],
+)
+def test_to_matrix(circuit: QCircuit, expected_matrix: Matrix):
```
hJaffaliColibritd, line 382: Should we add more complicated circuit with many more gates, for which we computed the matrix using Maple or Mathematica for instance ?

hJaffaliColibritd (changes requested):

mpqp/core/instruction/gates/controlled_gate.py
```diff
@@ -40,6 +43,87 @@ def __init__(
 
         Gate.__init__(self, targets, label)
 
+    def to_matrix(self, nb_qubits: int = 0) -> Matrix:
+        """
+        Constructs the matrix representation of a controlled gate.
+
+        Args:
+            nb_qubits: The total number of qubits in the system. If not provided,
```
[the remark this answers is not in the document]
JulienCalistoTD: it's not really the number of the QCircuit.  It's the number for the gate representation. But it's actually used to build the qcircuit, so maybe rename to be clearer?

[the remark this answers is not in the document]
Henri-ColibrITD: `desired_gate_size` ?

mpqp/core/instruction/gates/gate.py
```diff
@@ -40,6 +40,31 @@ def __init__(
 
     @abstractmethod
     def to_matrix(self) -> Matrix:
+        """Return the matricial semantics to this gate. With
+        considering potential column and row permutations needed if the targets.
```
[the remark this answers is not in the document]
Henri-ColibrITD: Indeed, this sentence is completely malformed ^^"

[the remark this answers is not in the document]
Henri-ColibrITD: I propose 
```
Considering connections' order and position, in contrast with :meth:`~Gate.to_canonical_matrix`.
```
(If you prefix the target of the link with `~`, the link text will only be the last component of the target. For example, :py:meth:`~Queue.Queue.get` will refer to Queue.Queue.get but only display get as the link text.

mpqp/core/instruction/gates/controlled_gate.py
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/core/instruction/gates/gate.py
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/core/instruction/gates/gate.py
```diff
@@ -77,6 +102,7 @@ def inverse(self) -> Gate:
             array([[1.-0.j, 0.-0.j],
                    [0.-0.j, 0.-1.j]])
 
+        # 3M-TODO: Multi qubits gate not handel by to_matrix
```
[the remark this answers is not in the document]
JulienCalistoTD: replace by "# 3M-TODO: test"

mpqp/core/instruction/gates/gate.py
```diff
@@ -103,6 +129,7 @@ def is_equivalent(self, other: Gate) -> bool:
             >>> X(0).is_equivalent(CustomGate(UnitaryMatrix(np.array([[0,1],[1,0]])),[1]))
             True
 
+        # 3M-TODO: Multi qubits gate not handel by to_matrix
```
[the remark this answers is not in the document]
JulienCalistoTD: replace by "# 3M-TODO: test"

mpqp/core/instruction/gates/gate_definition.py
JulienCalistoTD: done

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: what do you mean by not putting self.matrix directly ?

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: done

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: done

mpqp/noise/noise_model.py
JulienCalistoTD: it will print  AmplitudeDamping(0.3, targets=[0,1]):
targets = f", targets={self.targets}"                                      ->  targets=[0,1]
gates = f", gates={self.gates}" if len(self.gates) != 0 else "" -> ""
prob = f", prob={self.prob}" if self.prob != 1 else ""            -> ""
return f"{type(self).__name__}({self.gamma}{prob}{targets}{gates})" -> AmplitudeDamping({0.3}{""}  {, targets=[0,1]} {"}):
                                                                                                                                                    {prob}     {targets}    {gates}

mpqp/tools/circuit.py
JulienCalistoTD: done

mpqp/tools/generics.py
JulienCalistoTD: done

mpqp/tools/maths.py
JulienCalistoTD: We have a case where Expr.number is compared to an int, and the result was false:
0.00 == 0 → false.
So, first, we perform a comparison, and if it doesn't work, we compare the elements directly. It's basically the same thing that before.

mpqp/tools/generics.py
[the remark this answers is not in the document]
JulienCalistoTD: done

tests/core/test_circuit.py
JulienCalistoTD: for now the qcircuit .to_matrix is still the qiskit version

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: we created this "canonical matrix" thing to emphasize the difference, but maybe it is not needed indeed, we should have a quick discussion about this this afternoon

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: done

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: but is this mention even needed ? it feels like stating the obvious (multi qubits gates are not meant for single qubits circuits, of course...)

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: It's not necessarily the full size of the circuit, maybe we could rephrase this as 
```
Desired size of the gate (in term of qubits)
```

mpqp/noise/noise_model.py
Henri-ColibrITD: why did you remove this targets length check ? is seems useful to me
I don't think Hamza is right in his example, but I still think that for an empty target, you should not have in in the repr, were am I wrong ?

mpqp/tools/circuit.py
Henri-ColibrITD: we could also set the random seed for this test (or even for all tests) so that the result become stable

mpqp/tools/maths.py
Henri-ColibrITD: basically, what happens is that it tries to do type coercion, and it it fails it does the naive check, I think it's quite clean

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: yes i remove it

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: i put:  nb_qubits: The total number for qubits gate representation.

mpqp/noise/noise_model.py
JulienCalistoTD: true i discard the change. I think it was to specify the target for all the qubits, but we have to do it in __str__.

tests/core/test_circuit.py
hJaffaliColibritd: Ok makes sense then

mpqp/tools/maths.py
hJaffaliColibritd: Ok, thank you for the clarification

mpqp/core/instruction/gates/native_gates.py
hJaffaliColibritd: Sorry I was not clear, I mean in `to_matrix` we return `self.matrix`, and then in `to_canonical_matrix` we call the method `to_matrix`, my question was why we don't just directly return also `self.matrix` in `to_canonical_matrix`, instead of calling a method in a method (it is not super heavy in this case, but if we can simply avoid, it is better I think)

mpqp/noise/noise_model.py
hJaffaliColibritd: Indeed my example was not right.

mpqp/tools/circuit.py
hJaffaliColibritd: Should we create a task on trello ?

Henri-ColibrITD: done

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: done

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -150,7 +150,10 @@ def to_other_language(
     ):
         if qiskit_parameters is None:
             qiskit_parameters = set()
-        theta = float(self.theta) if self._numeric_parameters else self.theta
+        try:
```
henridb, line 153: since you changed this did you check if `_numeric_parameters` is still needed ?

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -587,8 +599,6 @@ def __init__(self, a: int, b: int):
 
         self.qlm_aqasm_keyword = "SWAP"
 
-        self.matrix = np.array([[1, 0, 0, 0], [0, 0, 1, 0], [0, 1, 0, 0], [0, 0, 0, 1]])
```
henridb, line 590: is it a mistake ?

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: I checked, it's not used anywhere, I removed it

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: I re-added it, explain me if it was a mistake

hJaffaliColibritd (approved):

PR #72, opened 2024-07-04
feat: amplitude damping noise channel

MoHermes:


mpqp/core/circuit.py
```diff
@@ -1054,6 +1054,7 @@ def pretty_print(self):
         qubits = set(range(self.size()[0]))
         if self.noises:
             for noise in self.noises:
+                # TODO: add support for other noise models, BitFlip, AmplitudeDamping, ....
```
hJaffaliColibritd, line 1057: This is not yet done right ? Also for BitFlip ?

mpqp/noise/noise_model.py
```diff
@@ -321,7 +323,116 @@ def to_kraus_representation(self) -> KrausRepresentation: ...
...
-    """3M-TODO"""
+    """Class representing the amplitude damping noise channel, which can model both
+    the standard and generalized amplitude damping processes. It can be applied
+    to a single qubit and depends on two parameters: the decay rate `gamma` and the
+    probability of excitation `prob`.
+
+    Args:
+        gamma: Decaying rate of the amplitude damping noise channel.
+        prob: Probability of excitation in the generalized amplitude damping noise channel
+            When set to 0, indicating standard amplitude damping.
```
hJaffaliColibritd, line 334: When set to 1

mpqp/noise/noise_model.py
```diff
@@ -321,7 +323,116 @@ def to_kraus_representation(self) -> KrausRepresentation: ...
...
+        NoiseModel:
+            AmplitudeDamping(gamma=0.2, prob=0, targets=[0])
+            AmplitudeDamping(gamma=0.4, prob=0.1, targets=[1, 2])
+
+    """
+
+    def __init__(
+        self,
+        gamma: float,
+        prob: float,
```
hJaffaliColibritd, line 365: We should put to optional and the default value to 1, so if the user wants the basic amplitude damping, he only inputs gamma, no need to touch prob

mpqp/noise/noise_model.py
```diff
@@ -321,7 +323,116 @@ def to_kraus_representation(self) -> KrausRepresentation: ...
...
+    def to_other_language(
+        self, language: Language = Language.QISKIT
+    ) -> BraketNoise | QLMNoise:
+        """See documentation of this method in abstract mother class :class:`NoiseModel`.
+
+        Args:
+            language: Enum representing the target language.
+
+        Examples:
+            >>> braket_ad = AmplitudeDamping(0.2, 0, [1]).to_other_language(Language.BRAKET)
```
hJaffaliColibritd, line 406: Put prob to 1 here to get the simple amplitude damping

mpqp/noise/noise_model.py
```diff
@@ -321,7 +323,116 @@ def to_kraus_representation(self) -> KrausRepresentation: ...
...
+            <class 'braket.circuits.noises.AmplitudeDamping'>
+            >>> braket_gad = AmplitudeDamping(0.15, 0.2, [0]).to_other_language(Language.BRAKET)
+            >>> braket_gad
+            GeneralizedAmplitudeDamping('gamma': 0.15, 'probability': 0.2, 'qubit_count': 1)
+            >>> type(braket_gad)
+            <class 'braket.circuits.noises.GeneralizedAmplitudeDamping'>
+
+        """
+        if language == Language.BRAKET:
+            if self.prob == 0:
```
hJaffaliColibritd, line 419: self.prob == 1

hJaffaliColibritd (changes requested):

hJaffaliColibritd:
Add some simple tests to check that the class can be instantiated and that the noisy simulation launches correctly.

mpqp/noise/noise_model.py
```diff
@@ -337,8 +337,134 @@ class PhaseFlip(NoiseModel):
...
+             ├───┤
+        q_1: ┤ H ├
+             ├───┤
+        q_2: ┤ H ├
+             └───┘
+        NoiseModel:
+            AmplitudeDamping(gamma=0.2, prob=0, targets=[0])
+            AmplitudeDamping(gamma=0.4, prob=0.1, targets=[1, 2])
+            AmplitudeDamping(gamma=0.1, prob=1, targets=[0, 1, 2])
+            AmplitudeDamping(gamma=0.7, prob=1, targets=[0, 1])
```
hJaffaliColibritd, line 360: I think the __repr__ can be simplified here. No need to put all the keywords, if prob=1, we can avoid writing it since it is the default value

mpqp/noise/noise_model.py
```diff
@@ -337,8 +337,134 @@ class PhaseFlip(NoiseModel):
...
+
+    """
+
+    def __init__(
+        self,
+        gamma: float,
+        prob: float = 1,
```
hJaffaliColibritd, line 367: You should use other type than ``float`` if you give int values (like 0 or 1) for prob. I would advise using Real (from numbers)

hJaffaliColibritd (changes requested):

mpqp/noise/noise_model.py
MoHermes: float works fine, and if you give 0 the error was coming from Braket when to_other_language, but with casting, it works fine

mpqp/noise/noise_model.py
MoHermes: This one already modified with the new __repr__, that was the old one, just will re-run the examples and all good

hJaffaliColibritd (changes requested):
In circuit.py, in pretty_print(), we only put the probability in the initialization of ``noise_info``. But this is not the only ``noise_info`` we have to print. For Depolarizing it is probability + the dimension if not equal to 1, for AmplitudeDamping it is gamma (first) and then the probability if it is not 1.
image

I suggest that we retrieve the initialization of ``noise_info`` directly from the NoiseModel calling maybe a method ``noise_model.pretty_info()``, that will return the noise_info, and this then is handled in each class separately.

The rest of the pretty print, especially the part related to targets should stay in circuit.py, because there we know the current number of qubits.

hJaffaliColibritd (approved):

Henri-ColibrITD (approved):

PR #75, opened 2024-08-06
Feat targets optional

JulienCalistoTD:


examples/scripts/observable_job.py
```diff
@@ -32,7 +32,7 @@
 # Constructing the circuit by adding gates and measurements
 circuit.add(H(0))
 circuit.add(Rx(1.76, 1))
-circuit.add(ExpectationMeasure([0, 1], observable=obs, shots=1000))
+circuit.add(ExpectationMeasure(obs, [0, 1], shots=1000))
```
Henri-ColibrITD, line 35: if targets are optional, should leave them here ?

Henri-ColibrITD: same question for all the following examples where targets involved span the whole circuit

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -194,60 +194,65 @@ class ExpectationMeasure(Measure):
...
-        super().__init__(targets, shots, label)
-        self.observable = observable
-        """See parameter description."""
-        # Raise an error if the number of target qubits does not match the size of the observable.
-        if self.nb_qubits != observable.nb_qubits:
-            raise NumberQubitsError(
-                f"{self.nb_qubits}, the number of target qubit(s) doesn't match"
-                f" {observable.nb_qubits}, the size of the observable"
-            )
+        if targets is None:
```
Henri-ColibrITD, line 204: in this case we don't define other attributes like `pre_measure` won't this pose problems at execution time ?

mpqp/noise/noise_model.py
```diff
@@ -344,6 +344,8 @@ def __init__(
             raise ValueError(
                 f"Invalid probability: {prob} but should be between 0 and 0.5"
             )
+        if targets is None:
```
Henri-ColibrITD, line 347: shouldn't this be dealt with and the `NoiseModel` level ?

Henri-ColibrITD (changes requested):
a few questions + no tests for measurement with empty targets ?

examples/scripts/observable_job.py
JulienCalistoTD: done

mpqp/core/instruction/measurement/expectation_value.py
JulienCalistoTD: done

mpqp/noise/noise_model.py
JulienCalistoTD: done

Henri-ColibrITD (approved):

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -204,16 +204,20 @@ def __init__(
...
+
+        if targets is None:
+            self.pre_measure = QCircuit(0)
+            return
+
         if self.nb_qubits != observable.nb_qubits:
             raise NumberQubitsError(
-                f"{self.nb_qubits}, the number of target qubit(s) doesn't match"
-                f" {observable.nb_qubits}, the size of the observable"
+                f"Target size {len(targets)} doesn't match observable size "
```
hJaffaliColibritd, line 214: Why do we use `len(targets)` instead of `self.nb_qubits` ? We already use the second in the `if`, and it would avoid calling `len` again

hJaffaliColibritd (changes requested):

mpqp/core/instruction/measurement/measure.py
```diff
@@ -42,14 +42,16 @@ class Measure(Instruction, ABC):
 
     def __init__(
         self,
-        targets: list[int],
+        targets: Optional[list[int]] = None,
         shots: int = 0,
```
hJaffaliColibritd, line 46: Shouldn't we put `shots: int = 1024` by default in `Measure` now that it is the case for both `BasisMeasure` and `ExpectationMeasure` ?

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -194,8 +194,8 @@ class ExpectationMeasure(Measure):
 
     def __init__(
         self,
-        targets: list[int],
         observable: Observable,
+        targets: Optional[list[int]] = None,
         shots: int = 0,
```
hJaffaliColibritd, line 199: Shouldn't we put `shots: int = 1024` like we did for `BasisMeasure` ?

mpqp/execution/runner.py
```diff
@@ -178,6 +178,8 @@ def _run_single(
          Error: None
 
     """
+    circuit = circuit.hard_copy()
```
hJaffaliColibritd, line 181: If we put a hard_copy of the circuit here, we will need to go into each provider file and remove it each time we call `hard_copy` on the circuit (to avoid a double call). This may result in a task in Trello since some are done in other branches.

tests/core/instruction/gates/test_gate.py
```diff
@@ -125,3 +126,21 @@ def test_add_product(g1: Gate, g2: Gate, result_matrix: npt.NDArray[np.complex64
...
 def test_sub_product(g1: Gate, g2: Gate, result_matrix: npt.NDArray[np.complex64]):
     assert np.allclose(g1.minus(g2).to_matrix(), result_matrix)
+
+
+@pytest.mark.parametrize(
+    "gate, targets",
+    [
+        (H.range(4), [0, 1, 2, 3, 4]),
+        (H.range(4, step=2), [0, 2, 4]),
+        (H.range(2, 4), [2, 3, 4]),
```
hJaffaliColibritd, line 136: It is said in the documentation that the `end` index is excluded, so we should have `[2,3]`in the targets, no ?

The tests are passing I guess because we iterate over the gates in the circuit, so we stop before addressing the remaining elements in `targets`.

tests/core/instruction/gates/test_gate.py
```diff
@@ -125,3 +126,21 @@ def test_add_product(g1: Gate, g2: Gate, result_matrix: npt.NDArray[np.complex64
...
+        (H.range(2, 4, 2), [2, 4]),
+        (Z.range(0, 4, 2), [0, 2, 3, 4]),
```
hJaffaliColibritd, line 138: If the step is 2, and the end is 4, we should have as targets `[0, 2]`, since the end is excluded, and 3 is not a multiple of 2.

The tests are passing I guess because we iterate over the gates in the circuit, so we stop before addressing the remaining elements in `targets`.

hJaffaliColibritd (changes requested):

mpqp/core/instruction/gates/gate.py
```diff
@@ -354,3 +354,22 @@ def __repr__(self) -> str:
...
     )
+
+    @classmethod
+    def range(cls, start_or_end: int, end: Optional[int] = None, step: int = 1):
+        """Apply the gate to a range of qubits.
+
+        Args:
+            start_or_end: If ``end`` is not defined, this value is treated as
+                the end value of the range, and the range starts from ``0``.
+                Otherwise, it is treated as the start value.
```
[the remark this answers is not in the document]
Henri-ColibrITD: u sure 'bout that buddy ?

mpqp/core/instruction/measurement/expectation_value.py
Henri-ColibrITD: We had a discussion about that: in basis measure, it is less useful to specify shots=0. Indeed, this would mean you are basically asking for a state vector job, which is equivalent to putting no measure. So in the case someone wants a state vector job, we would suggest putting no measure, rather that a basis measure with shots=0. On the other hand, for the expectation measure, this is not so true, so having the default being 0 is not crazy to me

Henri-ColibrITD: on the other hand, having different defaults for different measures might be weird, indeed

mpqp/core/instruction/gates/gate.py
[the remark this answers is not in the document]
hJaffaliColibritd: image

Both are correct apparently, "starting point" looks more common, "start point" more UK. We can ask our english referent ^^

tests/core/instruction/gates/test_gate.py
Henri-ColibrITD: fixed in dcd88

tests/core/instruction/gates/test_gate.py
Henri-ColibrITD: same

mpqp/core/instruction/gates/gate.py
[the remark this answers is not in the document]
Henri-ColibrITD: I'd say that using start instead of starting here is probably a bit less correct, but sticks closer to the variable name, it is also matching the "end point" part (which could also be "ending point" I guess)

mpqp/core/instruction/measurement/expectation_value.py
Henri-ColibrITD: For all measurements, say in the doc that shots=0 means no shot noise

Henri-ColibrITD (changes requested):

mpqp/execution/runner.py
hJaffaliColibritd: I tested this, and apparently, the circuit is not updated correctly
```python
>>> from mpqp.all import *
>>> c = QCircuit([H(0), CNOT(1,3), BasisMeasure()])
>>> c.nb_cbits

>>> c.hard_copy().nb_cbits
4
>>> print(run(c,ATOSDevice.MYQLM_PYLINALG)) # we run the circuit, to call _run_single, which should update the circuit
Result: None, ATOSDevice, MYQLM_PYLINALG
 Counts: [481, 0, 0, 0, 0, 0, 0, 0, 543, 0, 0, 0, 0, 0, 0, 0]
 Probabilities: [0.4697266, 0, 0, 0, 0, 0, 0, 0, 0.5302734, 0, 0, 0, 0, 0, 0, 0]
 Samples:
  State: 0000, Index: 0, Count: 481, Probability: 0.4697265625
  State: 1000, Index: 8, Count: 543, Probability: 0.5302734375
 Error: 0.015603954651445045

>>> c.nb_cbits

>>> c
QCircuit([H(0), CNOT(1,3), BasisMeasure(, shots=1024)], nb_qubits=4, nb_cbits=None, label="None")
```

We see that after execution, the `nb_cbits` of the circuit is not updated, nor the BasisMeasure,

mpqp/core/instruction/measurement/expectation_value.py
hJaffaliColibritd: I agree with this, we can let it as it is.

mpqp/execution/runner.py
JulienCalistoTD: It works well, it is not designed to update the circuits after a run. This applies to cases where we run the circuit and then change the number of qubits. However, I came up with a new idea for the instruction. Instead of providing an empty target list, we can directly specify the targets and introduce a new variable in the instruction, such as 'dynamic: bool = False'. When we update the number of qubits, we can check 'dynamic = True' and change the targets. So we don't have to use hard_copy . What do you think?

hJaffaliColibritd: I agree with idea, as discussed. It is fully implemented now, right?

mpqp/core/instruction/measurement/measure.py
```diff
@@ -42,16 +42,20 @@ class Measure(Instruction, ABC):
...
                 f"Negative number of shot makes no sense, given {shots}"
             )
+        if targets is None:
+            targets = []
         super().__init__(targets, label)
         self.shots = shots
         """See parameter description."""
         self.label = label
         """See parameter description."""
+        if targets == []:
```
hJaffaliColibritd, line 60: Is it better to put len(targets) == 0 ?

hJaffaliColibritd (approved):

mpqp/execution/runner.py
JulienCalistoTD: yes ready to review

PR #77, opened 2024-08-18
gates-repr

MoHermes:
fix wrong `__repr__` for all gates

### DIFF
--- a/mpqp/core/instruction/gates/native_gates.py
+++ b/mpqp/core/instruction/gates/native_gates.py
@@ -590,11 +590,9 @@ class SWAP(InvolutionGate, NoParameterGate):
         self.matrix = np.array([[1, 0, 0, 0], [0, 0, 1, 0], [0, 1, 0, 0], [0, 0, 0, 1]])
 
         super().__init__([a, b], "SWAP")
-        self.a = a
-        self.b = b
 
     def __repr__(self) -> str:
-        return f"{type(self).__name__}({self.a}, {self.b})"
+        return f"{type(self).__name__}({self.targets[0]}, {self.targets[1]})"
 
     nb_qubits = (  # pyright: ignore[reportAssignmentType,reportIncompatibleMethodOverride]
         2
--- a/mpqp/core/instruction/gates/native_gates.py
+++ b/mpqp/core/instruction/gates/native_gates.py
@@ -593,7 +593,7 @@ class SWAP(InvolutionGate, NoParameterGate):
         self.a = a
         self.b = b
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"{type(self).__name__}({self.a}, {self.b})"
 
     nb_qubits = (  # pyright: ignore[reportAssignmentType,reportIncompatibleMethodOverride]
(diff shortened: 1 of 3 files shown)


mpqp/core/instruction/gates/native_gates.py
```diff
@@ -902,6 +911,9 @@ def __init__(self, control: int, target: int):
     def to_matrix(self) -> Matrix:
         return np.array([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0]])
 
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}({self.controls[0]}, {self.targets[0]})"
```
hJaffaliColibritd, line 915: Don't you think that we can factorize that code for all controlled gates and have an automated code that handles single and multi-control (like for TOF) for the __repr__ ?
This would avoid replicating the same code for all controlled gates. The only particular case would be CRk, but it is already there so it can overcharge the one we would write for ControlledGate.

hJaffaliColibritd (changes requested):

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -588,7 +588,13 @@ def __init__(self, a: int, b: int):
         self.qlm_aqasm_keyword = "SWAP"
 
         self.matrix = np.array([[1, 0, 0, 0], [0, 0, 1, 0], [0, 1, 0, 0], [0, 0, 0, 1]])
+
         super().__init__([a, b], "SWAP")
+        self.a = a
+        self.b = b
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}({self.a}, {self.b})"
```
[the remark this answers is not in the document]
MoHermes: Yes correct, I was having it like this at the begining, but then changed to introduce more clarity when reading the `SWAP` class implementation, but indeed `self.targets` ensure consistency with parent class for any changes.

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: I actually don't think it's the case, because a controlled gate could have multiple control qubits. If we really wanted to factorize the behavior I guess we could have a generic implementation for `ControlledGate`s and add a class like `UniControlledGate` for gates with only one control qubit, but at this point I'm not sure enough gates are concerned by this to make it worth having this change. I'll let you convince me otherwise

Henri-ColibrITD: for demonstration purpose I did what I mentioned in the branch feat-gates-repr-facto-experience (commit 0e5f0f8), I'll let you be judge if you want it or not

hJaffaliColibritd: @Henri-ColibrITD @MoHermes I proposed another version that does not require the definition of another class but checks directly the `targets` and `controls` attributes to format correctly the `__repr__method ( see branch https://github.com/ColibrITD-SAS/mpqp/tree/feat-gates-repr-facto-2)

hJaffaliColibritd (approved):

PR #81, opened 2024-09-10
Feat Azure connetion and remote execution

JulienCalistoTD:


docs/resources/job-device_compat.csv
```diff
@@ -43,7 +43,17 @@ AWS      ,BRAKET_OQC_LUCY                   ,✓      ,             ,✓
...
+Google   ,IONQ_SIMULATOR                    ,✓      ,             ,           , 
+Google   ,IONQ_QPU                          ,✓      ,             ,           ,
+Azure    ,IONQ_SIMULATOR                    ,✓      ,             ,           ,
+Azure    ,IONQ_QPU                          ,✓      ,             ,           ,
+Azure    ,IONQ_QPU_ARIA_1                   ,✓      ,             ,           ,
+Azure    ,IONQ_QPU_ARIA_2                   ,✓      ,             ,           ,
+Azure    ,QUANTUM_SIM_H1_1                  ,✓      ,             ,           ,
+Azure    ,QUANTUM_SIM_H1_1SC                ,✓      ,             ,           ,
+Azure    ,QUANTUM_SIM_H1_1E                 ,✓      ,             ,           ,
+Azure    ,RIGETTI_SIM_QVM                   ,✓      ,             ,           ,
```
Henri-ColibrITD, line 57: were all these devices tested ?

mpqp/execution/connection/azure_connection.py
```diff
@@ -0,0 +1,126 @@
...
+    was_configured = get_env_variable(f"AZURE_CONFIGURED") == "True"
+
+    if was_configured:
+        decision = input(
+            f"AZURE account is already configured. Do you want to update it? [y/N]"
+        )
+        if decision.lower().strip() != "y":
+            return "Canceled.", []
+
+    resource_id = getpass(f"Enter your AZURE resource_id: ")
```
Henri-ColibrITD, line 28: in the doc (more specifically in getting-started#setup-remote-accesses), explain how to get these values

mpqp/execution/connection/env_manager.py
```diff
@@ -124,3 +127,37 @@ def save_env_variable(key: str, value: str) -> bool:
         return False
 
     return a
+
+
+
+def config_key(key_name: str, configuration_name: str, test_connection:  Callable[[str], bool]):
+    """
+    Configure a key by setting the API token.
+
```
Henri-ColibrITD, line 136: function arguments not documented

mpqp/execution/connection/ionq_connection.py
```diff
@@ -0,0 +1,65 @@
...
+    Returns:
+        tuple: A message indicating the result of the configuration and an empty list.
+    """
+    return config_key("IONQ_API_KEY", "IONQ", test_ionq_connection)
+
+
+def test_ionq_connection(key: Optional[str] = None) -> bool:
+    """
+    Test the connection to the IonQ service.
+
```
Henri-ColibrITD, line 20: func arg doc

mpqp/execution/devices.py
```diff
@@ -330,3 +330,36 @@ def is_processor(self) -> bool:
...
+    MICROSOFT_ESTIMATOR = "microsoft.estimator"
+
+    def is_remote(self):
+        return True
+
+    def is_gate_based(self) -> bool:
+        return True
+
+    def is_simulator(self) -> bool:
+        return True
```
Henri-ColibrITD, line 359: ionq qpus are simulators ?

mpqp/execution/devices.py
```diff
@@ -330,3 +330,36 @@ def is_processor(self) -> bool:
...
+
+    def is_noisy_simulator(self) -> bool:
```
Henri-ColibrITD, line 361: none of those support noise ?

mpqp/execution/providers/azure.py
```diff
@@ -0,0 +1,202 @@
...
+
+@typechecked
+def extract_result(
+    result: "MicrosoftEstimatorResult | QiskitResult",
+    job: Optional[Job],
+    device: AZUREDevice,
+) -> Result:
+    from qiskit.result import Result as QiskitResult
+
+    if isinstance(result, QiskitResult):
```
Henri-ColibrITD, line 85: can't we use the qiskit result extractor for this ?

mpqp/execution/providers/azure.py
```diff
@@ -0,0 +1,202 @@
...
+    """Retrieves from Azure remote platform and parse the result of the job_id
+    given in parameter. If the job is still running, we wait (blocking) until it
+    is ``DONE``.
+
+    Args:
+        job_id: Id of the remote Azure job.
+
+    Returns:
+        The result converted to our format.
+    """
```
Henri-ColibrITD, line 156: I would have imagined for this method something in the lines of
```py
job = get_jobs_by_id(job_id)
result = job.get_results()
return extract_result(result)
```
why is this not correct ?

mpqp/execution/providers/google.py
```diff
@@ -310,6 +328,18 @@ def extract_result_STATE_VECTOR(
     return Result(job, state_vector, 0, 0)
 
 
+def extract_result_OBSERVABLE_ideal2(
```
Henri-ColibrITD, line 331: why the 2 ?

mpqp/tools/display.py
```diff
@@ -72,61 +85,75 @@ def clean_1D_array(
...
-            )
-        )
-        for element in array
-    ]
+    cleaned_array = [format_element(element, round) for element in array]
     return "[" + ", ".join(map(str, cleaned_array)) + "]"
 
 
-def clean_matrix(matrix: Matrix):
+def clean_matrix(matrix: Matrix, round: int = 5, align: bool = True):
```
Henri-ColibrITD, line 108: nice additions !

Henri-ColibrITD (changes requested):
Very good work, a few questions to answer

mpqp/execution/connection/env_manager.py
JulienCalistoTD: done

mpqp/execution/connection/azure_connection.py
JulienCalistoTD: done

mpqp/execution/connection/ionq_connection.py
JulienCalistoTD: done

mpqp/execution/devices.py
JulienCalistoTD: done

mpqp/execution/devices.py
JulienCalistoTD: for now yes. we will fix it with noise for cirq

mpqp/execution/providers/azure.py
JulienCalistoTD: I have started working with this, but the result from a job id is not the same as the result of a run; it is not a Qiskit result.

mpqp/execution/providers/google.py
JulienCalistoTD: rename to extract_result_OBSERVABLE_processors

mpqp/execution/providers/azure.py
JulienCalistoTD: No, it is not exactly the same result, and it causes trouble.

docs/resources/job-device_compat.csv
JulienCalistoTD: done

Henri-ColibrITD:
tests needed

docs/resources/job-device_compat.csv
Henri-ColibrITD: the tests should come under the long flag (or maybe an expensive flag x) )

mpqp/execution/providers/azure.py
```diff
@@ -0,0 +1,202 @@
...
+    if TYPE_CHECKING:
+        assert isinstance(qiskit_circuit, QuantumCircuit)
+
+    qiskit_circuit = qiskit_circuit.reverse_bits()
+
+    backend_sim = get_azure_provider().get_backend(job.device.value)
+
+    assert isinstance(job.device, AZUREDevice)
+
+    # if job.job_type == JobType.STATE_VECTOR:
```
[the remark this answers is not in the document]
Henri-ColibrITD: Ok, I'll have a look in the coming days to see is I can do something about that

Henri-ColibrITD:
Still a few remarks to work on/answer

mpqp/execution/providers/azure.py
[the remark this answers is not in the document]
Henri-ColibrITD: I checked, and I came to the same conclusion as you

mpqp/execution/connection/azure_connection.py
Henri-ColibrITD, line 28: if we use getpass, we should explain that it's hidden (as we do for the other providers)

mpqp/execution/connection/azure_connection.py
```diff
@@ -0,0 +1,126 @@
...
+    if resource_id == "":
+        print(colored("Empty credentials", "red"))
+        getpass("Press 'Enter' to continue")
+        return "", []
+    Location = getpass(f"Enter your AZURE Location: ")
```
Henri-ColibrITD, line 33: maybe this doesn't need to be hidden ? I don't have strong opinion on the topic

mpqp/execution/connection/ionq_connection.py
```diff
@@ -0,0 +1,67 @@
...
+    ionq_api_key = get_env_variable("IONQ_API_KEY")
+    if ionq_api_key == "":
+        display = "Not configured"
+    else:
+        display = ionq_api_key[:5] + "*****"
+
+    return "   IONQ_API_KEY: " + display
+
+
+def get_ionq_job_ids() -> list[str]:
```
Henri-ColibrITD, line 55: add to "Remote devices handling" introduction in doc

mpqp/execution/providers/azure.py
Henri-ColibrITD: this is fucked up :') can you tell me more ?

Henri-ColibrITD (changes requested):
see the last few questions and we're gucci

docs/resources/job-device_compat.csv
JulienCalistoTD: long flag should be ok

mpqp/execution/providers/azure.py
JulienCalistoTD: for azure:
`data = [
                Sample(
                    bin_str="".join(map(str, state)),
                    nb_qubits=job.circuit.nb_qubits,
                    count=int(count),
                    probability=(
                        job_data.get("probabilities").get(state)
                        if "probabilities" in job_data
                        else None
                    ),
                )
                for (state, count) in job_data.get("counts").items()
            ]
`
for qiskit:
`data = [
                    Sample(
                        bin_str=item,
                        count=counts[item],
                        nb_qubits=job.circuit.nb_qubits,
                        probability=(
                            job_data.get("probabilities").get(item)
                            if "probabilities" in job_data
                            else None
                        ),
                    )
                    for item in counts
                ]
`

for both cases is a QiskitResult, but they are not the same elements, so it's a mess to make just one code.

tests/execution/test_result.py
```diff
@@ -92,11 +92,11 @@ def test_result_right_type(job_type: JobType, data: float | StateVector | list[S
...
+ Probabilities: [0.225, 0.37667, 0.01333, 0.385]
  Samples:
   State: 00, Index: 0, Count: 135, Probability: 0.225
-  State: 01, Index: 1, Count: 226, Probability: 0.37666666666666665
-  State: 10, Index: 2, Count: 8, Probability: 0.013333333333333334
+  State: 01, Index: 1, Count: 226, Probability: 0.37667
+  State: 10, Index: 2, Count: 8, Probability: 0.01333
   State: 11, Index: 3, Count: 231, Probability: 0.385
  Error: None""",
         ),
```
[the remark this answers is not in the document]
JulienCalistoTD: Would you like a flag for the remote simulator?

mpqp/execution/connection/azure_connection.py
JulienCalistoTD: replace by:
`"Enter your AZURE resource_id (hidden): "`

mpqp/execution/connection/azure_connection.py
JulienCalistoTD: replace by input()

tests/execution/test_result.py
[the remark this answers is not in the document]
Henri-ColibrITD: in --long should be fine (we have --long-local I think for long things that are specifically not remote)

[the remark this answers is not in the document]
hJaffaliColibritd: > in --long should be fine (we have --long-local I think for long things that are specifically not remote)

I confirm, we have that for QLM simus

docs/execution-extras.rst
```diff
@@ -72,31 +72,42 @@ __________
...
-.. _con-setup:
+IonQ
+^^^^
 
-Connection setup
-^^^^^^^^^^^^^^^^
+Connection
+__________
+
+.. automodule:: mpqp.execution.connection.ionq_connection
```
hJaffaliColibritd, line 88: We didn't add a section for ``Azure`` and ``mpqp.execution.connection.azure_connection", right ?

mpqp/execution/connection/azure_connection.py
```diff
@@ -0,0 +1,126 @@
...
+            save_env_variable(f"AZURE_RESOURCE_ID", resource_id)
+            save_env_variable(f"AZURE_LOCATION", Location)
+            save_env_variable(f"AZURE_CONFIGURED", "False")
+        getpass("Press 'Enter' to continue")
+        return "", []
+
+
+def get_azure_account_info() -> str:
+    """
+    Get the IonQ API key from the environment variables.
```
hJaffaliColibritd, line 54: Azure API key

mpqp/execution/connection/azure_connection.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Returns:
+        str: A string containing the IonQ API key.
```
hJaffaliColibritd, line 57: the Azure API key

mpqp/execution/connection/azure_connection.py
```diff
@@ -0,0 +1,126 @@
...
+    else:
+        display = azure_resource_id[:5] + "*****"
+    azure_location = get_env_variable("AZURE_LOCATION")
+
+    return "   AZURE_RESOURCE_ID: " + display + "\n   AZURE_LOCATION: " + azure_location
+
+
+def test_connection(resource_id: str, Location: str) -> bool:
+    """
+    Test the connection to the IonQ service.
```
hJaffaliColibritd, line 71: Azure service

mpqp/execution/providers/azure.py
```diff
@@ -0,0 +1,195 @@
...
+                        if "probabilities" in job_data
+                        else None
+                    ),
+                )
+                for (state, count) in job_data.get("counts").items()
+            ]
+            return Result(job, data, None, job.measure.shots)
+        else:
+            raise NotImplementedError(f"{job.job_type} not handled.")
+    else:
```
hJaffaliColibritd, line 132: if the result is a `MicrosoftEstimatorResult`, there is only one jobtype possible I suppose? (observable one ?)

If this assomption is correct, it looks like there is an error in the Result object you return line 135. You give first the job, and then 0 for the `data` argument of the constructor, and then result.data() for the `errors`. Is there an error here?

mpqp/execution/providers/google.py
```diff
@@ -237,8 +241,22 @@ def run_local_processor(job: Job) -> Result:
...
+        from cirq.ops.pauli_string import PauliString as Cirq_PauliString
+
+        assert isinstance(job.measure, ExpectationMeasure)
+
+        cirq_obs = job.measure.observable.to_other_language(
+            language=Language.CIRQ, circuit=cirq_circuit
+        )
+        assert type(cirq_obs) == Cirq_PauliSum or type(cirq_obs) == Cirq_PauliString
+
+        shots = 1000 if job.measure.shots == 0 else job.measure.shots
```
hJaffaliColibritd, line 254: Usually when we put `job.measure.shots` to 0, we want the exact expectation value, why here do we put manually shots to 1000?
If it is not possible to have the theoretical expectation value, then we should have return an error (DeviceJobIncompatibleError) earlier (in the check job device comptability for instance) or then implement it, right?

mpqp/execution/providers/ibm.py
```diff
@@ -651,11 +651,17 @@ def extract_result(
             elif job.job_type == JobType.SAMPLE:
                 assert job.measure is not None
                 counts = result.get_counts(0)
+                job_data = result.data()
                 data = [
                     Sample(
                         bin_str=item,
                         count=counts[item],
                         nb_qubits=job.circuit.nb_qubits,
+                        probability=(
```
hJaffaliColibritd, line 660: You dont' need to fill the probability argument if you put the counts and the number of qubits, this will be done directly in the Sample constructor.

Update: I just saw that you modified the way we create samples, and allowed to directly set both if we have them. So you can ignore my remark. The only small problem we can have, is if the counts and the probabilities are not coherent (when retrieved from the provider), we will have sample that have inconsistency between counts and probas (and here recomputing one from the other will avoid that).

mpqp/tools/display.py
```diff
@@ -62,8 +67,35 @@ def _remove_unnecessary_decimals(val: np.float32 | int) -> np.float32 | int:
     return val
 
 
+def format_element(element: Union[int, float, complex | Expr], round: int = 5) -> str:
```
hJaffaliColibritd, line 70: Doc and example ?

mpqp/execution/providers/azure.py
JulienCalistoTD: in azure.py:
```py 
88 def get_result_from_azure_job_id:
101 	 result = job.get_results()
```
return :
```py
typing.Any
Results dictionary with histogram shots, or raw results if not a json object. 
-> (dict[Any, Any] | Any | dict[Unknown, Unknown] | Unknown)
```

so if we pass it to extract_result
we need to pass result to `Any`

we can do it but we will lost the type_check

JulienCalistoTD: And if I put the extraction code in `extract_resul`, there's no need to create `get_result_from_azure_job_id`.

mpqp/execution/connection/azure_connection.py
JulienCalistoTD: done

mpqp/execution/connection/azure_connection.py
JulienCalistoTD: done

mpqp/execution/connection/azure_connection.py
JulienCalistoTD: done

docs/execution-extras.rst
JulienCalistoTD: done

tests/execution/test_result.py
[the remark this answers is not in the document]
JulienCalistoTD: add -long-cost

mpqp/execution/providers/azure.py
JulienCalistoTD: `MicrosoftEstimatorResult` is not a normal result, nor a sample, nor an observable, nor a state vector. It just gives a message about the estimate. At the moment, we're putting the message in error -> str. But I think we'll have to create a different type of result.

mpqp/execution/providers/ibm.py
JulienCalistoTD: Yes, that's exactly why I created this feature. We can have the probability we should have and the result we get. I don't think it's a problem to have inconsistencies, is it?

mpqp/execution/result.py
```diff
@@ -161,7 +161,10 @@ def __init__(
                     )
 
     def __str__(self):
-        return f"State: {self.bin_str}, Index: {self.index}, Count: {self.count}, Probability: {self.probability}"
+        return (
+            f"State: {self.bin_str}, Index: {self.index}, Count: {self.count}"
+            + f", Probability: {np.round(self.probability, 5) if self.probability is not None else None}"
```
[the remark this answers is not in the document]
JulienCalistoTD: We create the probability or count in the `Result` and not in `Sample`, so the `Sample` can have `self.probability` None. We want to create the probability in the result because we need all the samples.
```py
self._probabilities = np.array(counts, dtype=float) / self.shots
for sample in self._samples:
   sample.probability = self._probabilities[sample.index]
```

mpqp/execution/providers/google.py
JulienCalistoTD: done

mpqp/tools/display.py
JulienCalistoTD: done

tests/execution/providers/test_azure.py
```diff
@@ -0,0 +1,72 @@
+"""Add ``--long`` to the cli args to run this test with credits (disabled by default)"""
```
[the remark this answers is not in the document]
JulienCalistoTD: done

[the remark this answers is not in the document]
JulienCalistoTD: done

hJaffaliColibritd:
Did you try to setup the credentials with the script on a docker ?

docs/getting-started.rst
```diff
@@ -58,7 +58,7 @@ Setup remote accesses
 
 Installing MPQP gives you access to ``setup_connections`` a script facilitating
 the setting up of remote QPU connections. The three supported providers (qiskit,
```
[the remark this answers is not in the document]
hJaffaliColibritd: Indeed Google is not a provider, however from a pure mpqp point of view, we call `GOOGLEDevice.IONQ_QPU`.
I agree to continue the discussion on another branch dedicated to doc.

mpqp/execution/providers/ibm.py
hJaffaliColibritd: If when I print the samples from a result, and I have counts and probability that are not coherent, I think that indeed it is a problem.
I would say that if we allow to input both counts and probas in the creation of sample, we have to check that both are consistent (because we supposed we cannot trust the result coming from the providers), and raise an error if they are not.
But this check will be equivalent to recompute the probas from the counts, or the counts from the probas, so that is why I think at the end that we have to chose one over both inputs (I will go with counts to ensure exactitude of the result, compared to probas that can be rounded).

mpqp/execution/result.py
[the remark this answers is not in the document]
hJaffaliColibritd: Indeed. See discussion above, but I still think that we never have `self.probability` to `None`, even in your new version of the code, right?
If the user only put the counts, then the probas will be computed by us.
If the user inputs both counts and probas, then the probas will be filled directly from there.
Correct me if I am wrong.

mpqp/execution/providers/azure.py
hJaffaliColibritd: > And if I put the extraction code in `extract_result`, there's no need to create `get_result_from_azure_job_id` except for passing the job_id.

About this comment, even if we put this extraction here in `extract_result` the method of getting the result from azure job id is still useful, because it is the one that the user will use directly by just inputting the id. In our building of the execution process, and user never calls by himself the method extract_result, that is why I prefer to move everything there.

hJaffaliColibritd: > in azure.py:
> 
> ```python
> 88 def get_result_from_azure_job_id:
> 101 	 result = job.get_results()
> ```
> 
> return :
> 
> ```python
> typing.Any
> Results dictionary with histogram shots, or raw results if not a json object. 
> -> (dict[Any, Any] | Any | dict[Unknown, Unknown] | Unknown)
> ```
> 
> so if we pass it to extract_result we need to pass result to `Any`
> 
> we can do it but we will lost the type_check

Ummm this is a bit annoying that the providers return `Any` but we can't really do nothing on that part.

However, a possible escape is to define a function `extract_remote_result` or something like this specifically for this case, move it to `azure.py`, and then call this function from here. What do you think?

JulienCalistoTD: Ok but if it is juste for `get_result_from_azure_job_id`, Is it still useful to make a second function if it's only used ones?

hJaffaliColibritd (approved):

PR #82, opened 2024-09-17
feat: mpqp to qasm

JulienCalistoTD:


JulienCalistoTD:
wait for qcircuit to matrix for test random circuit.
+ write doc

mpqp/qasm/mpqp_to_qasm.py
```diff
@@ -0,0 +1,218 @@
...
+            + " "
+            + _handle_connection(instruction)
+            + ";"
+        )
+
+        return "\n" + qasm_str
+
+    elif isinstance(instruction, Gate):
+        instruction_str = (
+            instruction.qasm2_gate  # pyright: ignore[reportAttributeAccessIssue]
```
Henri-ColibrITD, line 94: this smells like architecture mishap, if we say that a gate "has to" have a
`qasm2_game`, then we wouldn't need this type ignore

mpqp/qasm/mpqp_to_qasm.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+        if instruction_str is None:
```
Henri-ColibrITD, line 96: and it would also mean that instruction_str cannot be `None`

mpqp/qasm/mpqp_to_qasm.py
```diff
@@ -0,0 +1,218 @@
...
+def _float_to_str(f: float) -> str:
+    if f.is_integer():
+        return str(int(f))
+    elif f % np.pi == 0:
+        return f"{int(f/np.pi)}*pi" if f != np.pi else "pi"
+    else:
+        return f"pi/{int(1 / f * np.pi)}" if (np.pi * (1 / f)).is_integer() else str(f)
+
+
+def _handle_rotation(gate: ParametrizedGate) -> str:
```
Henri-ColibrITD, line 49: can't this be done at the gate level ? something like `Gate.to_other_language`
would eliminate the need for a separate function here

Henri-ColibrITD: This appliues to many of the private functions here

mpqp/qasm/mpqp_to_qasm.py
```diff
@@ -0,0 +1,218 @@
...
+                if previous is None or not isinstance(instruction, type(previous)):
+                    if previous:
+                        qasm_str += _simplify_instruction(previous, targets)
+                        targets = {i: 0 for i in range(circuit_copy.nb_qubits)}
+                    previous = instruction
+                else:
+                    if (
+                        isinstance(instruction, ParametrizedGate)
+                        and instruction.parameters
+                        != previous.parameters  # pyright: ignore[reportAttributeAccessIssue]
```
Henri-ColibrITD, line 200: I would do the export in two steps, first an export using instructions'
`to_other_language` methods, then a simplification step, with simple rewriting step

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -296,6 +296,8 @@ def qiskit_gate(cls):
 
         return IGate
 
+    qasm2_gate = "id"
```
Henri-ColibrITD, line 299: if we move this to the instr constructor, we could have this in the `uper().__init__`

tests/qasm/test_mpqp_to_qasm.py
```diff
@@ -0,0 +1,92 @@
...
+            TOF([0, 1], 2),
+        ],
+        [Barrier()],
+        [X(0), Y(1), BasisMeasure([0])],
+        [CNOT(1, 0), CNOT(0, 1)],
+    ],
+)
+def test_mpqp_to_qasm(instructions: list[Instruction]):
+    circuit = QCircuit(instructions)
+    assert mpqp_to_qasm2(circuit, False) == circuit.to_qasm2()
```
Henri-ColibrITD, line 86: if we remove the `to_quasm2` and `mpqp_to_qasm2`, this could become 
```py
circ.to_other_language(Language.QISKIT).qasm() == circ.to_other_language(Language.QASM2)
```

mpqp/core/instruction/barrier.py
```diff
@@ -31,3 +31,6 @@ def to_other_language(
             from qiskit.circuit.library import Barrier as QiskitBarrier
 
             return QiskitBarrier(self.size)
+        elif language == Language.QASM2:
+            qubits = ",".join([f"q[{j}]" for j in self.targets])
+            return "\nbarrier " + qubits + ";"
```
Henri-ColibrITD, line 36: u sure about the line break ? I would not put it here and instead add it at the circuit export level. Something like
```py
def to_other_language(self, language):
    if language == Language.QASM2:
        return "\n".join(self.instructions.to_other_language(Language.QASM2)
```

mpqp/core/instruction/breakpoint.py
```diff
@@ -30,3 +34,13 @@ def __repr__(self) -> str:
             f"Breakpoint(targets={self.targets}, draw_circuit={self.draw_circuit},"
             f" enabled={self.enabled}, label={self.label})"
         )
+
+    def to_other_language(
```
Henri-ColibrITD, line 38: I would `pass` and filter them out at the circuit level

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -267,7 +267,9 @@ def to_other_language(
         self,
         language: Language = Language.QISKIT,
         qiskit_parameters: Optional[set["Parameter"]] = None,
-    ) -> None:
+    ) -> None | str:
```
Henri-ColibrITD, line 270: same as for breakpoints

mpqp/core/instruction/breakpoint.py
JulienCalistoTD: done

mpqp/core/instruction/measurement/expectation_value.py
JulienCalistoTD: done

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -177,6 +237,31 @@ def to_other_language(
...
                 )
             return self.braket_gate(theta)
+        if language == Language.QASM2:
+            from mpqp.qasm.mpqp_to_qasm import float_to_qasm_str
+
+            instruction_str = self.qasm2_gate
+            if isinstance(self, (Rk, CRk)):
+                instruction_str += (
+                    "("
+                    + float_to_qasm_str(2 * np.pi / (2 ** float(self.parameters[0])))
```
Henri-ColibrITD, line 247: couldn't this be `float_to_qasm_str(self.theta)` instead ?

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -413,6 +529,10 @@ def qiskit_gate(cls):
 
         return ZGate
 
+    @classproperty
```
Henri-ColibrITD, line 532: for this I think you could have `qasm2_gate = "z"` instead of a `classproperty`, we used class properties on the other ones to avoid the import in the class directly, because this would affect the library load time

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -177,6 +237,31 @@ def to_other_language(
                     "export, this feature is coming very soon!"
                 )
             return self.braket_gate(theta)
+        if language == Language.QASM2:
+            from mpqp.qasm.mpqp_to_qasm import float_to_qasm_str
+
+            instruction_str = self.qasm2_gate
+            if isinstance(self, (Rk, CRk)):
```
Henri-ColibrITD, line 244: if we have a special case for those gates, should this go directly in the body of these classes ?

mpqp/core/instruction/measurement/basis_measure.py
```diff
@@ -83,6 +83,17 @@ def to_other_language(
                 return Measure()
             else:
                 raise NotImplementedError(f"{type(self.basis)} not handled")
+        if language == Language.QASM2:
```
Henri-ColibrITD, line 86: I think we should check that the computational basis is used

Henri-ColibrITD (changes requested):
a few tweaks and we're good

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: done

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: done

mpqp/core/instruction/measurement/basis_measure.py
JulienCalistoTD: done

mpqp/qasm/mpqp_to_qasm.py
```diff
@@ -0,0 +1,154 @@
...
+                    targets = {i: 0 for i in range(circuit.nb_qubits)}
+                    previous = instruction
+                for target in instruction.targets:
+                    targets[target] += 1
+            else:
+                if previous:
+                    qasm_str += _simplify_instruction(previous, targets)
+                    previous = None
+                    targets = {i: 0 for i in range(circuit.nb_qubits)}
+                if isinstance(instruction, CustomGate):
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: I agree, but the problem is that you have to redo to_other_language every time, with all the other languages.

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: I replaced it with :
`instruction_str += f"({float_to_qasm_str(2 * np.pi / (2 ** float(self.k)))})"`
but I cannot replace it with theta, since theta is :
`return pi / 2 ** (self.k - 1)`

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -741,7 +847,19 @@ def to_other_language(
...
+            instruction_str += (
+                "("
+                + ",".join(float_to_qasm_str(float(param)) for param in self.parameters)
+                + ")"
+            )
+            qubits = ",".join([f"q[{j}]" for j in self.targets])
+
+            return instruction_str + " " + qubits + ";"
         else:
             raise NotImplementedError(f"Error: {language} is not supported")
```
hJaffaliColibritd, line 877: Here a comment on the translation to OpenQASM3 for a circuit containing the gate `U`.

We pass through openqasm2 that will transform the gate `U`to an openQasm2 `u` (with potentially handling of the global phase).

However, when we will translate to openqasm3, we will consider that this is a gate `u` (without nowing it was coming from a `U` gate in mpqp that is the same as the openqasm3 `U`), and so it will translate it to the openqasm3 gate `u3`, while it should have been the native keyword `U` in openQASM3, and this will generate for sure a global phase problem.

This today may affect `AWSDevice`s since we use that pipeline to translate. We should in the future do the translation directly to OpenQASM3 if we can.

One can verify this statement with the following code: 
```Python
>>> from mpqp.all import *
>>> c = QCircuit([U(1,2,3,0)])
>>> print(c.to_other_language(Language.QASM3))
OPENQASM 3.0;
include "stdgates.inc";

qubit[1] q;
u3(1,2,3) q[0];
```

If one checks the definition of `u3` (and its matrix) and compare with the mpqp `U`, he will witness the problem.

mpqp/core/circuit.py
```diff
@@ -288,6 +288,7 @@ def _update_targets_components(self, components: Instruction | NoiseModel):
 
         if isinstance(components, Barrier):
             components.size = self.nb_qubits
+            components.targets = [i for i in range(self.nb_qubits)]
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/core/circuit.py
```diff
@@ -1040,107 +1048,20 @@ def to_other_language(
...
-
+        elif language == Language.QASM2:
+            from mpqp.qasm.mpqp_to_qasm import mpqp_to_qasm2
+
+            qasm_str, gphase = mpqp_to_qasm2(self)
+            self.gphase = gphase
+            return qasm_str
+        elif language == Language.QASM3:
+            qasm2_code = self.to_other_language(Language.QASM2)
+            assert isinstance(qasm2_code, str)
```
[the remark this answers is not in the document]
JulienCalistoTD: yes, but this is the case for all assert. Do you want to put a `if TYPE_CHEKING` for each assert ?

mpqp/core/circuit.py
```diff
@@ -860,7 +861,7 @@ def without_noises(self) -> QCircuit:
 
     def to_other_language(
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -82,12 +83,55 @@ def to_other_language(
         self,
         language: Language = Language.QISKIT,
         qiskit_parameters: Optional[set["Parameter"]] = None,
+        qcircuit: Optional["QCircuit"] = None,
```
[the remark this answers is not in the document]
JulienCalistoTD: the doc is eritate from the other to_other_language

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -82,12 +83,55 @@ def to_other_language(
...
+            if qiskit_parameters is None:
+                qiskit_parameters = set()
+            return QiskitOperator(self.matrix)
+        elif language == Language.QASM2:
+            from mpqp.tools.circuit import replace_custom_gate
+            from qiskit.quantum_info.operators import Operator as QiskitOperator
+            from qiskit import QuantumCircuit, qasm2
+
+            if qcircuit:
+                nb_qubits = qcircuit.nb_qubits
```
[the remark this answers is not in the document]
JulienCalistoTD: to_other_language eritate so that the parameter is the same, and we already have qcircuit

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -82,12 +83,55 @@ def to_other_language(
...
+            qiskit_circ = QuantumCircuit(nb_qubits)
+
+            qiskit_circ.unitary(
+                QiskitOperator(self.matrix).to_instruction(),
+                list(reversed(self.targets)),  # dang qiskit qubits order
+                self.label,
+            )
+            filtered_qasm = ""
+            final_gphase = 0
+            for instruction in qiskit_circ.data:
```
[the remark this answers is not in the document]
JulienCalistoTD: do you propose to get instructions from `qiskit_circ.data[0]` ? If that's the case, I think it's better to use a for

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -95,6 +95,10 @@ class NativeGate(Gate, SimpleClassReprABC):
         label: Label used to identify the gate.
     """
 
+    qasm2_gate: str
```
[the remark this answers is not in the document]
JulienCalistoTD: There are the same, but if there is only one gate that is not the same, we have to redeclare the `qasm2_gate`, and it is not consistent to call `qiskit_string` to get the qasm2 gate.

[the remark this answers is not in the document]
JulienCalistoTD: Perhaps we can put `qasm2_gate` in the `Gate` class which calls `qiskit_string` and if not, we can redefine `qasm2_gate` for the specifique gate.

mpqp/core/languages.py
```diff
@@ -24,3 +24,5 @@ class Language(Enum):
     MY_QLM = 1
     BRAKET = 2
     CIRQ = 3
+    QASM2 = 4
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/qasm/mpqp_to_qasm.py
```diff
@@ -0,0 +1,176 @@
...
+    from mpqp.core.circuit import QCircuit
+
+from mpqp.core.instruction.gates import *
+from mpqp.core.instruction.gates.gate import SingleQubitGate
+from mpqp.core.languages import Language
+from mpqp.core.instruction.breakpoint import Breakpoint
+from mpqp.core.instruction.measurement import ExpectationMeasure, BasisMeasure
+
+
+def float_to_qasm_str(f: float) -> str:
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/qasm/mpqp_to_qasm.py
```diff
@@ -0,0 +1,176 @@
...
+    if f.is_integer():
+        return str(int(f))
+    elif f % np.pi == 0:
+        return f"{int(f/np.pi)}*pi" if f != np.pi else "pi"
+    else:
+        return f"pi/{int(1 / f * np.pi)}" if (np.pi * (1 / f)).is_integer() else str(f)
+
+
+def _simplify_instruction(
```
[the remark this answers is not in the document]
JulienCalistoTD: TypeChecked done

mpqp/qasm/mpqp_to_qasm.py
```diff
@@ -0,0 +1,176 @@
...
+                        final_str += f"c[{key}],"
+                        c_targets[key] -= 1
+                final_str = final_str[:-1]
+            final_str += ";"
+
+    return final_str
+
+
+def mpqp_to_qasm2(circuit: QCircuit, simplify: bool = False) -> tuple[str, float]:
+    """Converts an :class:`~mpqp.core.circuit.QCircuit` object into a string in
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/qasm/mpqp_to_qasm.py
```diff
@@ -0,0 +1,176 @@
...
+                    if c_target != 0:
+                        final_str += f"c[{key}],"
+                        c_targets[key] -= 1
+                final_str = final_str[:-1]
+            final_str += ";"
+
+    return final_str
+
+
+def mpqp_to_qasm2(circuit: QCircuit, simplify: bool = False) -> tuple[str, float]:
```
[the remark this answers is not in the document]
JulienCalistoTD: done

tests/qasm/test_mpqp_to_qasm.py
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/core/instruction/measurement/basis_measure.py
Henri-ColibrITD: I would do something like 
```py
return measure.pre_measure().to_other_language(language) + f"\nmeasure q[{target}] c[{i}];"
```

Alternatively, if we already cover the basis change where this function is called, we should raise a warning in case the user calls this function themself. Something like `"measure in non computational basis, don't forget about adding the basis change circuit before measuring"`. And potentially a way to filter out this warning when we call the function ourselves.

mpqp/core/circuit.py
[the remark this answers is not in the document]
Henri-ColibrITD: that's usually what we do, to avoid having the performance penalty of the asserts

mpqp/core/instruction/gates/custom_gate.py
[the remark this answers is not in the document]
Henri-ColibrITD: indeed, the doc is inherited from `Instruction`, but it doesn't have this `qcircuit` parameter, so I think what Hamza meant is that you have to override it to add the documentation of this parameter

mpqp/core/instruction/gates/custom_gate.py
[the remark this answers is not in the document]
Henri-ColibrITD: he's right, we pass an object much bigger than needed (even though we only pass the reference, so it's not such a big deal, but I'd still pass `nb_qubits` only, if it's the only thing you need)

mpqp/core/instruction/gates/custom_gate.py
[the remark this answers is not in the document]
Henri-ColibrITD: maybe
```py
qiskit_gate = QiskitOperator(self.matrix).to_instruction()
qiskit_circ.unitary(
    qiskit_gate,
    list(reversed(self.targets)),  # dang qiskit qubits order
    self.label,
)
circuit, gphase = replace_custom_gate(qiskit_gate, nb_qubits)
...
```

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: what about defining it for most cases here, but overloading it for `Rk` et al ?

tests/qasm/test_mpqp_to_qasm.py
```diff
@@ -0,0 +1,334 @@
...
+        ]
+    ],
+)
+def test_mpqp_to_qasm_custom_gate(instructions: list[Instruction]):
+    circuit = QCircuit(instructions)
+    from qiskit import qasm2, QuantumCircuit
+
+    qiskit_circuit = circuit.to_other_language(Language.QISKIT)
+    assert isinstance(qiskit_circuit, QuantumCircuit)
+    str_qiskit_circuit = remove_user_gates(qasm2.dumps(qiskit_circuit), True)
```
Henri-ColibrITD, line 113: this looks like circular logic to me, aren't we doing this exact same line in `o_other_language(Language.QASM2)` ?

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -267,14 +273,15 @@ def add_std_lib():
...
-There is a phase e^(i(a+c)/2) difference between U(a,b,c) gate in 2.0 and 3.0.
-We handled that for you by adding the extra phase at the right place. 
-Be careful if you want to create a control gate from this circuit/gate, the
-phase can become non-global.""",
-            OpenQASMTranslationWarning,
-        )
+        if translation_warning:
+            warn(
+                """
+    There is a phase e^(i(a+c)/2) difference between U(a,b,c) gate in 2.0 and 3.0.
```
Henri-ColibrITD, line 279: u sure about the tabs here ?

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: Done: I overloading to_other_language for `Rk` and `CRk`. And i create a task to make the overload function more flexible to avoid rewriting all the functions for the specific case.

mpqp/core/circuit.py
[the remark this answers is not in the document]
JulienCalistoTD: Done: check for all `asserts isinstance()` to put a `if TYPE_CHECKING:`

mpqp/core/instruction/gates/custom_gate.py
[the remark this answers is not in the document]
JulienCalistoTD: we need that the instructions are instantiated in the circuit. so we can't just pass `qiskit_gate`

mpqp/core/instruction/gates/native_gates.py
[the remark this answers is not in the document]
JulienCalistoTD: done

tests/qasm/test_mpqp_to_qasm.py
JulienCalistoTD: No, we use the mpqp_to_qasm2 method. But if you prefer, we can hard-code the test ?

mpqp/core/instruction/gates/native_gates.py
Henri-ColibrITD: you can call `super().to_other_language` for the cases in which the overloading is not needed

mpqp/qasm/open_qasm_2_and_3.py
JulienCalistoTD: done: remove tab

mpqp/core/instruction/gates/custom_gate.py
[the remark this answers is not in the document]
Henri-ColibrITD: haaa, in that case indeed both your solution and Hamza's one are the best suited, on one hand you're is "cleaner" (no special case) but Hamza's is clearer (in reality we do not iterate, since we know there is only one item), I'll let you discuss with him what to chose.

mpqp/core/instruction/gates/custom_gate.py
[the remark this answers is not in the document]
JulienCalistoTD: in the end we don't need `qcircuit` so i removed it

mpqp/core/instruction/gates/custom_gate.py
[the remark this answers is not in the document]
JulienCalistoTD: `qcircuit` removed

tests/qasm/test_mpqp_to_qasm.py
JulienCalistoTD: done

mpqp/core/instruction/gates/custom_gate.py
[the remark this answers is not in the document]
JulienCalistoTD: done

PR #90, opened 2024-10-02
Fix custom gate

hJaffaliColibritd:


tests/core/instruction/gates/test_custom_gate.py
```diff
@@ -0,0 +1,166 @@
...
+    # building the expected state vector
+    exp_state_vector = m[:, 0]
+    for _ in range(0, t):
+        exp_state_vector = np.kron(np.array([1, 0]), exp_state_vector)
+    for _ in range(t + n_gate, n_circ):
+        exp_state_vector = np.kron(exp_state_vector, np.array([1, 0]))
+
+    execution_ibm_statevector = run(c, IBMDevice.AER_SIMULATOR).state_vector
+    execution_aws_statevector = run(c, AWSDevice.BRAKET_LOCAL_SIMULATOR).state_vector
+    execution_qlm_statevector = run(c, ATOSDevice.MYQLM_PYLINALG).state_vector
```
JulienCalistoTD, line 48: We might be able to do a batch result

tests/core/instruction/gates/test_gate.py
```diff
@@ -71,6 +71,7 @@ def test_power(gate: Gate, pow: float, result_matrix: npt.NDArray[np.complex64])
     assert np.allclose(gate.power(pow).to_matrix(), result_matrix)
 
 
+@pytest.mark.skip
```
JulienCalistoTD, line 74: why skip this test?

mpqp/core/instruction/gates/gate_definition.py
```diff
@@ -123,15 +123,27 @@ def __init__(self, definition: Matrix, disable_symbol_warn: bool = False):
                 "Matrices defining gates have to be unitary. It is not the case"
                 f" for\n{definition}"
             )
+        elif not is_power_of_two(definition.shape[0]):
```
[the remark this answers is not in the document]
hJaffaliColibritd: Is unitary only check if the operator is unitary, not if it is suitable for qubits (which is only the case if the square matrix has dimension that is power of two), so I would say we need this extra check

tests/core/instruction/gates/test_gate.py
hJaffaliColibritd: Because the feature of tensor product between gates is not fully implemented, so the test is ready but will not succeed, so we skip it

mpqp/core/instruction/gates/gate_definition.py
```diff
@@ -123,15 +123,27 @@ def __init__(self, definition: Matrix, disable_symbol_warn: bool = False):
                 "Matrices defining gates have to be unitary. It is not the case"
```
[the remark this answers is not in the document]
hJaffaliColibritd: You we replace the symbolic by specific values, and if the matrix is not unitary it means that the symbolic one was also not ?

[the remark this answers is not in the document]
JulienCalistoTD: No, sometimes we have an expression without a variable, so can do a evalf. But if we have a variable, i think it's too difficult to know if it's a unitary.

JulienCalistoTD (approved):

mpqp/core/instruction/gates/gate_definition.py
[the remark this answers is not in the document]
hJaffaliColibritd: Maybe a way to check is to compute the doc product between each line/column and then ask sympy to simplify it, and check that it is always zero. We also have to check that each line/colmun forms a vector of norm 1, which I think can also be done with symbolic calculus. Imma add a task on Trello.

mpqp/core/circuit.py
```diff
@@ -133,6 +132,11 @@ def __init__(
         self._nb_qubits: int
         """Number of qubits of the circuit."""
 
+        self.gphase: Optional[float] = None
```
Henri-ColibrITD, line 135: wouldn't it be simpler if it was 1 by default ? and not optional ?

mpqp/core/circuit.py
```diff
@@ -913,14 +918,14 @@ def to_other_language(
                 cargs = []
 
                 if isinstance(instruction, CustomGate):
+                    # We reverse the target qubits to pass the unitary in the usual qubit ordering
+                    reversed_target = instruction.targets
+                    reversed_target.reverse()
                     new_circ.unitary(
                         instruction.to_other_language(),
-                        instruction.targets,
+                        reversed_target,
```
Henri-ColibrITD, line 926:
```py
reversed(instruction.targets)
```

mpqp/core/circuit.py
```diff
@@ -1045,16 +1052,78 @@ def to_qasm2(self) -> str:
...
         qiskit_circ = self.subs({}, remove_symbolic=True).to_other_language(
             Language.QISKIT
         )
         if TYPE_CHECKING:
             assert isinstance(qiskit_circ, QuantumCircuit)
 
-        from qiskit import qasm2
+        qiskit_circ.reverse_bits()
+
+        if all([not isinstance(inst, CustomGate) for inst in self.instructions]):
```
Henri-ColibrITD, line 1108: why is this needed ? as I see it the following loop will cover this case, and it doesn't seem to save a lot of time

mpqp/core/circuit.py
hJaffaliColibritd: Indeed, we can do that. However, `reversed(...)` returns an iterator, is `list(reversed(instructions.targets))` a suitable solution?

mpqp/core/circuit.py
hJaffaliColibritd: Yes we can indeed, then we check if it is different from 1, we multiply the state_vector by it at the result time, if 1 we do nothing.

hJaffaliColibritd: Done

mpqp/core/circuit.py
hJaffaliColibritd: Done

tests/core/instruction/gates/test_gate.py
Henri-ColibrITD: I implemented the tensor product i think, so we can run this test now

Henri-ColibrITD (changes requested):
see questions

mpqp/core/circuit.py
Henri-ColibrITD: the key point is that the two loops are not nested but one after the other, so it's `n+n` and not `n*n`, so I'd say it's not a big deal, but if you prefer to keep the shortcut I'm fine with it

hJaffaliColibritd: I introduced a boolean variable to check if wether we encountered a CustomGate in the circuit or not, and then dump the original or new qiskit circuit.

mpqp/execution/result.py
```diff
@@ -258,6 +258,9 @@ def __init__(
                 )
             else:
                 self._state_vector = data
+                if job.circuit.gphase is not None:
```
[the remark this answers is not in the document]
hJaffaliColibritd: Done

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -12,19 +17,49 @@
...
+
+    Note:
+        For the moment, only ordered and contiguous target qubits are allowed when instantiating a CustomGate.
+
     """
 
     def __init__(
         self, definition: UnitaryMatrix, targets: list[int], label: Optional[str] = None
     ):
         self.matrix = definition.matrix
```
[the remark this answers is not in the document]
hJaffaliColibritd: We can indeed. To limit the influence on the rest of the code, I propose to define an attribute `self.unitary` and put the definition, and then define a property `self.matrix` that returns the matrix of the unitary.

This may however make pyright scream a bit if he can't understand that the type of self.matrix is Matrix, so will require to add some type cheking.

I let you confirm the idea before I implement.

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -12,19 +17,49 @@
...
 
     Args:
-        definition: The matrix (this is the only way supported to now) semantics of the gate.
+        definition: The UnitaryMatrix (only way supported for now) description of the gate.
         targets: The qubits on which the gate operates.
         label: The label of the gate. Defaults to None.
+
+    Raises:
+        ValueError: the target qubits must be contiguous and in order, and must match the size of the UnitaryMatrix
+
```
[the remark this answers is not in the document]
hJaffaliColibritd: Not mandatory, we just need to re-use the work you did to generate any matrix from a gate, playing with the targets etc.

However, I think this should be done on another branch, after merging this one, because the work may not be straightfoward, when dealing with qiskit qubit ordering and change in the size of the CustomGate.

I create a task in Trello.

mpqp/execution/result.py
```diff
@@ -258,6 +258,9 @@ def __init__(
                 )
             else:
                 self._state_vector = data
+                if job.circuit.gphase != 0:
```
Henri-ColibrITD, line 261: is this check really needed ?

mpqp/core/circuit.py
Henri-ColibrITD: I removed this variable (with your permission) because it doesn't really has any role in the result

mpqp/core/instruction/gates/custom_gate.py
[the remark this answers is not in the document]
Henri-ColibrITD: meh, not sure it's important enough to spend that energy. It's just that if one day we add other ways to define gates, it may be annoying to handle.

Maybe just a task on trello ?

Henri-ColibrITD (approved):
a last comment to resolve but overall I'm fine with the PR !

mpqp/execution/result.py
hJaffaliColibritd: In case there is no custom gate (which I assume will be the most common usage), I think checking that the phase is null is less costful than calling `np.exp` independently of the phase.

PR #95, opened 2024-10-16
Phase Damping noise model

MoHermes:


examples/notebooks/6_Noise_Simulation.ipynb
```diff
@@ -606,16 +716,16 @@
...
    "cell_type": "code",
-   "execution_count": 21,
+   "execution_count": 24,
    "metadata": {},
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "QuantumError([(<qiskit.circuit.quantumcircuit.QuantumCircuit object at 0x7f6f97b34a90>, 1.0)])"
+       "QuantumError([(<qiskit.circuit.quantumcircuit.QuantumCircuit object at 0x7f20042ae880>, 1.0)])"
```
hJaffaliColibritd, line 725: We should avoid printing directly the error, because the __repr__ of Qiskit is not clean, and shows the address in memory of the object. I would print something else that is still interesting

examples/notebooks/6_Noise_Simulation.ipynb
```diff
@@ -655,16 +790,16 @@
...
    "cell_type": "code",
-   "execution_count": 24,
+   "execution_count": 28,
    "metadata": {},
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "QuantumError([(<qiskit.circuit.quantumcircuit.QuantumCircuit object at 0x7f6f8cc57eb0>, 1.0)])"
+       "QuantumError([(<qiskit.circuit.quantumcircuit.QuantumCircuit object at 0x7f2047dea850>, 1.0)])"
```
hJaffaliColibritd, line 799: Same remark as before about showing __repr__ of qiskit objects

examples/notebooks/6_Noise_Simulation.ipynb
```diff
@@ -673,6 +808,205 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 4- PhaseDamping noise model\n",
+    "\n",
+    "We can define the ``PhaseDamping`` noise model by providing the decoherence rate ``gamma``, which represents the probability of phase error occurring during the quantum operation. ``PhaseDamping`` noise model introduces phase errors into a quantum system caused by interactions with the environment, leading to decoherence but leaving the energy of the system unchanged.\n",
+    "\n",
+    "Similar to other noise models, One can target all qubits of a circuit or just select specific ones. If no targets are specified, the noise model will be applied to all qubits in the circuit. By default, the parameter gates is an empty list."
+   ]
```
hJaffaliColibritd, line 845: More precisely, by default the parameter gates is None, and then is turned into an empty list, but not sure those details are important from a user point of view

examples/notebooks/6_Noise_Simulation.ipynb
```diff
@@ -673,6 +808,205 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 35,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "QuantumError([(<qiskit.circuit.quantumcircuit.QuantumCircuit object at 0x7f2047ddce50>, 1.0)])"
```
hJaffaliColibritd, line 951: same comment about the memory adress that will change from a run to another

hJaffaliColibritd (changes requested):

examples/notebooks/6_Noise_Simulation.ipynb
MoHermes: Will remove the info about the gates from it? what do you think?

hJaffaliColibritd: you can maybe just say that if we don't fill the parameter 'gates', then all gates in the circuit are affected by the noise

hJaffaliColibritd (approved):

Henri-ColibrITD (approved):

Henri-ColibrITD:
I "fixed" the reprs, is my change correct ?

PR #98, opened 2024-10-21
Fix random seed

hJaffaliColibritd:


mpqp/tools/circuit.py
```diff
@@ -27,16 +25,17 @@
...
     """This function creates a QCircuit with a specified number of qubits and gates.
     The gates are chosen randomly from the provided list of native gate classes.
 
     args:
         nb_qubits : Number of qubits in the circuit.
         gate_classes : List of native gate classes to use in the circuit.
-        nb_gates : Number of gates to add to the circuit. Defaults to a random
-         integer between 5 and 10.
+        nb_gates : Number of gates to add to the circuit.
+        seed: Seed used to control the random generation of the circuit.
```
hJaffaliColibritd, line 38: Maybe precise the behavior when the optional seed is not set (random uniform int ?)

mpqp/tools/maths.py
```diff
@@ -229,7 +237,7 @@ def exp(angle: Expr | Complex) -> sp.Expr | complex:
 
 
 def rand_orthogonal_matrix(
-    size: int, seed: Optional[int] = None
+    size: int, seed: Optional[Union[int, np.random.Generator]] = None
```
hJaffaliColibritd, line 240: same thing, inform what happens when seed is not entered

mpqp/tools/maths.py
```diff
@@ -307,47 +351,71 @@ def rand_unitary_2x2_matrix() -> npt.NDArray[np.complex64]:
...
-    return reduce(np.kron, [rand_unitary_2x2_matrix() for _ in range(nb_qubits - 1)])
+    if seed is None:
+        rng = np.random.default_rng()
+    elif isinstance(seed, np.random.Generator):
+        rng = seed
+    else:
+        rng = np.random.default_rng(seed)
+
+    return reduce(
+        np.kron, [rand_unitary_2x2_matrix(seed=rng) for _ in range(nb_qubits - 1)]
```
hJaffaliColibritd, line 385: here we have a problem. If the user gives a seed for the product of local unitaries, he doesn't necessarly expected that all the 2x2 unitarties will have the same seed and will be identical.

I suggest that we find a way to decompose uniquely the seed given in parameter to multiple seed that will be passed one by one for each rand_unitary 2x2 matrix generation. (maybe use prime decomposition of the integer, and put 1 as a seed if we don't have enough factors)

tests/core/instruction/gates/test_custom_gate.py
```diff
@@ -10,8 +11,8 @@
     ATOSDevice,
     AvailableDevice,
     AWSDevice,
-    IBMDevice,
     GOOGLEDevice,
+    IBMDevice,
```
hJaffaliColibritd, line 15: why moving IBM after ? was it doen by isort ?

mpqp/tools/maths.py
MoHermes: But here Hamza, they are not the same!
The second specifies that the variable can hold either an int, a np.random.Generator (a generator for random numbers from the NumPy library), or be None.
?

hJaffaliColibritd: Ok, ut the question is the same, what happens if the parameter is None ?

tests/core/instruction/gates/test_custom_gate.py
MoHermes: Yes.

conftest.py
```diff
@@ -1,10 +1,28 @@
...
+        help="Set a global random seed for tests (default is None for random behavior).",
+    )
+
+
+@pytest.fixture
+def global_seed(request: pytest.FixtureRequest):
+    seed = request.config.getoption("--seed")
+
+    if seed is None:
+        print("\nNo seed is provided, using the default random behavior.")
```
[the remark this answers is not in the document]
MoHermes: Then ok

mpqp/tools/maths.py
MoHermes: I think here we could implment seed decomposition using incrementing values, so the overall seed generation is controlled by the initial user provided seed

mpqp/tools/maths.py
hJaffaliColibritd: Then this should appear in the documentation so the user knows. This is the same remark as above, and everywhere else we have ``seed`` appearing in the doc string

mpqp/tools/maths.py
```diff
@@ -147,7 +147,15 @@ def closest_unitary(matrix: Matrix):
         Return U as a numpy matrix.
 
     Example:
-        # TODO : fill examples
+        >>> m = np.array([[1, 2], [3, 4]])
+        >>> is_unitary(m)
+        False
+        >>> u = closest_unitary(m)
+        >>> u
+        array([[-0.51449576,  0.85749293],
```
hJaffaliColibritd, line 155: use pretty_print ?

mpqp/tools/maths.py
```diff
@@ -229,75 +237,117 @@ def exp(angle: Expr | Complex) -> sp.Expr | complex:
...
 
     Args:
         size: Size (number of columns) of the square matrix to generate.
+        seed: Seed used to control the random generation of the matrix.
 
     Returns:
         A random Clifford matrix.
 
     Examples:
         >>> rand_clifford_matrix(2) # doctest: +SKIP
```
hJaffaliColibritd, line 290: use pprint() ?

JulienCalistoTD:
I have a suggestion regarding the use of seeds during tests. The purpose of using a seed, is to allow us to reproduce a failing test with the same conditions. When we run pytest without any specific options, we want to know which tests fail and with what seed they were executed. However, if the seed isn't displayed, it becomes difficult to reproduce the test. So, wouldn’t it be helpful when we run pytest without option to use a random seed and print it?

mpqp/tools/circuit.py
```diff
@@ -46,25 +48,54 @@ def random_circuit(
 
     Examples:
         >>> random_circuit([U, TOF], 3) # doctest: +SKIP
```
Henri-ColibrITD, line 50: shouldn't we remove the `# doctest: +SKIP` here ?

mpqp/tools/maths.py
Henri-ColibrITD: meh

mpqp/tools/circuit.py
Henri-ColibrITD: done

PR #99, opened 2024-10-21
Feat braket cost estimation

hJaffaliColibritd:


mpqp/execution/providers/aws.py
```diff
@@ -319,3 +319,47 @@ def get_result_from_aws_task_arn(task_arn: str) -> Result:
     device = AWSDevice.from_arn(device_arn)
 
     return extract_result(result, None, device)
+
+
+def estimate_cost_single_job(
+    job: Job, hybrid_iterations: int = 1, estimated_time: int = 3
+) -> float:
```
Henri-ColibrITD, line 326: what about some documentation ? I'd also add the unit for the time in the variable name

mpqp/execution/providers/aws.py
```diff
@@ -319,3 +319,47 @@ def get_result_from_aws_task_arn(task_arn: str) -> Result:
...
+
+    if TYPE_CHECKING:
```
Henri-ColibrITD, line 328: here I would put an actual guardrail in case user try to call this function for non `AWSDevice`s, something like
```py
is not isinstance(job.device, AWSDevice):
    raise JobDeviceMismatchError(...)
```

mpqp/execution/providers/aws.py
```diff
@@ -319,3 +319,47 @@ def get_result_from_aws_task_arn(task_arn: str) -> Result:
...
+            if "sv1" in job.device.value or "dm1" in job.device.value:
+                minute_cost = 0.075
+            elif "tn1" in job.device.value:
+                minute_cost = 0.275
+            else:
+                raise ValueError
+            return minute_cost * max(estimated_time / 60, 3 / 60)
+        else:
+            if TYPE_CHECKING:
+                assert job.measure is not None
```
Henri-ColibrITD, line 342: same here, I would put a true type check

Henri-ColibrITD (changes requested):
see comments

mpqp/execution/providers/aws.py
hJaffaliColibritd: Done

mpqp/execution/providers/aws.py
hJaffaliColibritd: Done

mpqp/execution/providers/aws.py
hJaffaliColibritd: Really done this time

Henri-ColibrITD (approved):

PR #101, opened 2024-10-21
Feat custom basis

hJaffaliColibritd:


mpqp/core/instruction/measurement/basis.py
```diff
@@ -76,7 +83,7 @@ def __init__(
         ):
             raise ValueError("The given basis is not orthogonal")
 
-        self.nb_qubits = nb_qubits
+        self.nb_qubits: int = nb_qubits
```
Henri-ColibrITD, line 86: why is this needed ?

mpqp/core/circuit.py
```diff
@@ -940,12 +940,7 @@ def to_other_language(
                     qargs = instruction.controls + instruction.targets
                 elif isinstance(instruction, Gate):
                     qargs = instruction.targets
-                elif isinstance(instruction, BasisMeasure) and isinstance(
-                    instruction.basis, ComputationalBasis
-                ):
-                    # TODO for custom basis, check if something should be
```
Henri-ColibrITD, line 946: I would remove this only once the new logic to replace it is in place, is it ?

mpqp/core/instruction/measurement/basis.py
```diff
@@ -171,6 +190,11 @@ def set_size(self, nb_qubits: int):
         ]
         self.nb_qubits = nb_qubits
 
+    def to_computational(self):
```
Henri-ColibrITD, line 193: I think this will be needed on a per basis level, so I wouldn't put it here

Henri-ColibrITD:
note: so far the tests only mention Hadamard basis I believe, it that's indeed the case we should add tests with fully custom basis

mpqp/core/instruction/measurement/basis_measure.py
```diff
@@ -97,12 +101,12 @@ def __repr__(self) -> str:
             options += (
                 f", basis={self.basis}"
                 if targets and options
-                else f"basis={self.shots}"
+                else f", basis={self.basis}"
```
Henri-ColibrITD, line 104: wow, the basis=shots thing before was weird ! but shouldn't we put the shots anyway still ?

tests/core/instruction/measurement/test_basis.py
```diff
@@ -104,3 +108,21 @@ def test_basis_implementations(
...
+    custom_basis = Basis(basis_vectors)
+
+    circuit = QCircuit([H(0), CNOT(0, 1), SWAP(1, 2), Z(2)])
+    circuit.add(BasisMeasure([0, 1, 2], basis=custom_basis, shots=1024))
+
+    res = _run_single(circuit, IBMDevice.AER_SIMULATOR, {})
+    actual_probabilities = {sample.index: sample.probability for sample in res.samples}
+
+    assert all(
+        abs((actual_probabilities.get(int(index)) or 0) - expected_prob) < 0.02
```
Henri-ColibrITD, line 126: I'm sure this test count be done better, this seems very complicated for something that just check the equality between two vectors

mpqp/core/instruction/measurement/basis.py
```diff
@@ -190,13 +258,18 @@ class HadamardBasis(VariableSizeBasis):
     """
```
Henri-ColibrITD, line 258: same here: I would put an example showcasing the symbols for Hadamard (-+)

mpqp/core/instruction/measurement/basis.py
```diff
@@ -45,15 +47,18 @@ def __init__(
         self,
         basis_vectors: list[npt.NDArray[np.complex64]],
         nb_qubits: Optional[int] = None,
+        symbols: Optional[tuple[str, str]] = None,
```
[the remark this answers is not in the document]
Henri-ColibrITD: something like `↑` and `↓`

tests/core/instruction/measurement/test_basis.py
Henri-ColibrITD: done

mpqp/core/instruction/measurement/basis_measure.py
Henri-ColibrITD: fixed

tests/core/instruction/measurement/test_basis.py
```diff
@@ -104,3 +108,21 @@ def test_basis_implementations(
     b.pretty_print()
     captured = capsys.readouterr()
     assert captured.out == result_pp
+
+
+def test_run_with_custom_basis():
```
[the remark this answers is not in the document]
MoHermes: Done.

mpqp/core/instruction/measurement/basis.py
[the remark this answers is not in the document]
MoHermes: Yes, I added it

mpqp/core/instruction/measurement/basis.py
MoHermes: Done.

Henri-ColibrITD (approved):

hJaffaliColibritd:
```Python
>>> from mpqp.all import *
>>> u1 = np.array([np.sqrt(3)/2, 1/2])
>>> u2 = np.array([-1/2, np.sqrt(3)/2])
>>> b = Basis([u1,u2])
>>> c = QCircuit([X(0), X(0), BasisMeasure(basis=b)])
...
ValueError: A `BasisMeasure` with a non variable sized basis cannot be dynamic.
```

Here the number of qubits is inferred from the size of the basis. The number of qubits of the basis here is exactly the size of the circuit, meaning that there is no problem here of not putting the targets, since it will be `list(range(nb_qubits))`. Maybe we only raise an error when the size of the basis is strictly lower than the size of the circuit (which is equal to the size of the dynamic BasisMeasure when the size is set)?

hJaffaliColibritd:
```Python
>>> from mpqp.all import *
>>> u1 = np.array([np.sqrt(3)/2, 1/2])
>>> u2 = np.array([-1/2, np.sqrt(3)/2])
>>> b = Basis([u1,u2])
>>> c = QCircuit([X(0), X(0), BasisMeasure([0],basis=b)])
>>> print(run(c,ATOSDevice.MYQLM_PYLINALG))
...
NotImplementedError: Does not handle other basis than the ComputationalBasis for the moment
>>> print(run(c,GOOGLEDevice.CIRQ_LOCAL_SIMULATOR))
...
NotImplementedError: Does not handle other basis than the ComputationalBasis for the moment
```

I think we should update the execution workflow for those two AvailableDevices (and maybe also azure.py, I didn't check).

hJaffaliColibritd:
```Python
>>> from mpqp.all import *
>>> u1 = np.array([np.sqrt(3)/2, 1/2])
>>> u2 = np.array([-1/2, np.sqrt(3)/2])
>>> b = Basis([u1,u2])
>>> c = QCircuit([X(0), BasisMeasure([0],basis=b, shots=0)])
>>> print(run(c,[IBMDevice.AER_SIMULATOR,ATOSDevice.MYQLM_PYLINALG,AWSDevice.BRAKET_LOCAL_SIMULATOR,GOOGLEDevice.CIRQ_LOCAL_SIMULATOR]))
BatchResult: 4 results
Result: circuit 1, IBMDevice, AER_SIMULATOR
 State vector: [-0.5, 0.86603]
 Probabilities: [0.25, 0.75]
 Number of qubits: 1
Result: circuit 1, ATOSDevice, MYQLM_PYLINALG
 State vector: [-0.5, 0.86603]
 Probabilities: [0.25, 0.75]
 Number of qubits: 1
Result: circuit 1, AWSDevice, BRAKET_LOCAL_SIMULATOR
 State vector: [-0.5, 0.86603]
 Probabilities: [0.25, 0.75]
 Number of qubits: 1
Result: circuit 1, GOOGLEDevice, CIRQ_LOCAL_SIMULATOR
 State vector: [-1, 0]
 Probabilities: [1, 0]
 Number of qubits: 1
```

All the state vectors are correct except the one returned by `GOOGLEDevice.CIRQ_LOCAL_SIMULATOR`. Most of the time (i suppose 75% of the time) it gives the other state vector `State vector: [0, 1]`. Looks like it projects in the computational basis the rotated vector (while here we don't need to measure and project).

Henri-ColibrITD:
> ```python
> >>> from mpqp.all import *
> >>> u1 = np.array([np.sqrt(3)/2, 1/2])
> >>> u2 = np.array([-1/2, np.sqrt(3)/2])
> >>> b = Basis([u1,u2])
> >>> c = QCircuit([X(0), X(0), BasisMeasure(basis=b)])
> ...
> ValueError: A `BasisMeasure` with a non variable sized basis cannot be dynamic.
> ```
> 
> Here the number of qubits is inferred from the size of the basis. The number of qubits of the basis here is exactly the size of the circuit, meaning that there is no problem here of not putting the targets, since it will be `list(range(nb_qubits))`. Maybe we only raise an error when the size of the basis is strictly lower than the size of the circuit (which is equal to the size of the dynamic BasisMeasure when the size is set)?

good point

mpqp/core/instruction/measurement/basis.py
```diff
@@ -33,27 +35,41 @@ class Basis:
...
+          State: ↓↓↑, Index: 6, Count: 5064, Probability: 0.5064
+         Error: None
 
     """
 
     def __init__(
         self,
         basis_vectors: list[npt.NDArray[np.complex64]],
         nb_qubits: Optional[int] = None,
+        symbols: Optional[tuple[str, str]] = None,
```
hJaffaliColibritd, line 61: I think we should also give the possibility to the user to precise the symbols associated with each basis vector, instead of only allowing for pair single qubit symbols. Indeed, not all basis are build using a tensor products of 1-qubit vectors.

So for instance if I have a 2-qubit basis b={u1, u2, u3, u4} that cannot be expressed as {v1v1, v1v2, v2v1, v2v2} we should be able pass all the 2^n symbols for each basis state, so they can appear in the sampling result.

Henri-ColibrITD: makes sense, it would be mutually exclusive with symbols (both cannot be expressed together), something like `basis_vectors_labels` ?

hJaffaliColibritd: Yes indeed 👍

mpqp/core/instruction/measurement/basis_measure.py
```diff
@@ -79,30 +88,32 @@ def to_other_language(
...
+                if len(options) != 0 or len(targets) != 0
+                else f"basis={self.basis}"
             )
         if self.label is not None:
             options += (
                 f", label={self.label}"
-                if targets and options
-                else f"label={self.shots}"
+                if len(options) != 0 or len(targets) != 0
+                else f"label={self.label}"
```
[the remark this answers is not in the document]
Henri-ColibrITD: It's a tricky bit of logic: the chore part of it is: we should add a coma if things were written before. So to do that, at each step we have to check if there are things before or not.
Maybe an example can be useful to understand this ? let's say `target == []` but `basis != ComputationalBasis` and `label != ""`. In that case the first if will make it so there is no comma before the `basis`, but one is added before the `label`. Now, you can play with the different possible combinations, to see that the two `if` are indeed required. (We could even have a separate function that does something like `add_comma_if_needed`, but it would be so niche that I don't think it's worth it)

mpqp/core/instruction/measurement/basis.py
```diff
@@ -71,7 +87,7 @@ def __init__(
             raise ValueError("All vectors of the given basis are not normalized")
         m = np.array([vector for vector in basis_vectors])
         if not matrix_eq(
-            m.transpose().dot(m),
+            m.transpose().dot(m), #fixme
```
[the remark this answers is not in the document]
Henri-ColibrITD: good catch, it should be fixed in bb17fef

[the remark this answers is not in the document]
Henri-ColibrITD: note: there was a few errors in the commit, it's fixed in the following ones, but having a look at it will already give you a good indication of the fix

mpqp/core/instruction/measurement/basis.py
Henri-ColibrITD: done in [b30b039](https://github.com/ColibrITD-SAS/mpqp/pull/101/commits/b30b0394b8b5ce1d925380b7f827a29c5ce9e92a)

mpqp/core/circuit.py
```diff
@@ -26,13 +26,6 @@
 from typing import TYPE_CHECKING, Iterable, Optional, Sequence, Type
 from warnings import warn
```
Henri-ColibrITD, line 27: To continue Hamza's discussion about the problem with cirq, here are minimal examples to show the problem:
```py
>>> c = QCircuit([BasisMeasure(basis=Basis([np.array([0,1]), np.array([1,0])]), shots=0)], nb_qubits=1)
>>> run(c, GOOGLEDevice.CIRQ_LOCAL_SIMULATOR).state_vector
StateVector([-1, 0])
>>> run(c, IBMDevice.AER_SIMULATOR).state_vector
StateVector([0, 1])
```
while 
```py
>>> c = QCircuit([BasisMeasure(basis=Basis([np.array([1,0]), np.array([0,1])]), shots=0)], nb_qubits=1)
>>> run(c, GOOGLEDevice.CIRQ_LOCAL_SIMULATOR).state_vector
StateVector([1, 0])
>>> run(c, IBMDevice.AER_SIMULATOR).state_vector
StateVector([1, 0])
```

Henri-ColibrITD:
```py
>>> res = run(c, GOOGLEDevice.CIRQ_LOCAL_SIMULATOR)
>>> res.job.circuit.gphase
3.141592653589793
```

Henri-ColibrITD: It looks like the pre_measure circuit adds a phase, which makes sense, but then I don't get why the result is on the 1st basis vector and not the second one, could it be related to qiskit's qubits ordering deal ?

MoHermes: Based on cirq return, I think it has a different internal convention for ordering the states or a global phase that could affect the outcome??? normalization?, phase conventions?, .... investigating!!
While IBM uses the standard computational basis ordering `|0>`, `|1>`

mpqp/core/instruction/measurement/basis.py
```diff
@@ -96,17 +137,62 @@ def pretty_print(self):
...
+    """
 
     @abstractmethod
-    def __init__(self, nb_qubits: Optional[int] = None):
-        super().__init__([], nb_qubits)
-        pass
+    def __init__(
+        self, nb_qubits: Optional[int] = None, symbols: Optional[tuple[str, str]] = None
+    ):
+        if symbols is None:
```
hJaffaliColibritd, line 191: This is done already in the `super().__init__(...)`, so maybe we can remove those lines ?

mpqp/core/instruction/measurement/basis_measure.py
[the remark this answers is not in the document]
hJaffaliColibritd: Ok my bad, my mistake comes from the fact that I missed the coma while I was reading. Everything makes sense. I resolve both comments

mpqp/execution/runner.py
```diff
@@ -112,13 +112,12 @@ def generate_job(
     elif nb_meas == 1:
         measurement = m_list[0]
         if isinstance(measurement, BasisMeasure):
-            # TODO: handle other basis by adding the right rotation (change
-            # of basis) before measuring in the computational basis
-            # Muhammad: circuit.add(CustomGate(UnitaryMatrix(change_of_basis_inverse)))
+            modified_circuit = circuit.without_measurements() + measurement.pre_measure
```
hJaffaliColibritd, line 115: According to the discussion we had last time, we said that it would be more relevant to let the pre_measure handling at the provider part right?
In the sense that when the user wants to retrieve the circuit from the job (in the local case), the pre_measure doesn't appear. We also said that it can impact noisy simulations (where we expect that only the circuit gates are affected by noise, not the pre-measure we add for basis change that is just an ergonomic feature).

In any case, if we choose to do it here, we have to remove the specific treatment we do for some providers (for example line 144 in google.py, this can be the source of the problems we face with Cirq i think).

mpqp/core/instruction/measurement/basis.py
```diff
@@ -33,27 +35,48 @@ class Basis:
...
-        # basis, the symbols will be '+' and '-'. If the user wants '↑' and '↓'
-        # for his custom basis, when we print samples we would have something
-        # like:
-        # State: ↑↑↓, Index: 1, Count: 512, Probability: 0.512
+        if symbols is not None and basis_vectors_labels is not None:
+            raise ValueError(
+                "You can only specify either symbols or basis_vectors_labels, not both."
+            )
+
+        if symbols is None:
```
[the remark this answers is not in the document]
Henri-ColibrITD: good catch, you could use GitHub's `suggestion` feature for tiny changes like this one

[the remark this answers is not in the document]
Henri-ColibrITD: but in fact we always first check that `basis_vectors_labels` is not `None`, so this is not that relevant. We could add this check but it is rather insignificant (convince me otherwise :p )

mpqp/core/instruction/measurement/basis.py
```diff
@@ -63,24 +86,42 @@ def __init__(
...
         if any(abs(np.linalg.norm(vector) - 1) > atol for vector in basis_vectors):
-            raise ValueError("All vectors of the given basis are not normalized")
-        m = np.array([vector for vector in basis_vectors])
-        if not matrix_eq(
-            m.transpose().dot(m),
-            np.eye(len(basis_vectors)),  # pyright: ignore[reportArgumentType]
-        ):
-            raise ValueError("The given basis is not orthogonal")
+            raise ValueError("Some vectors of the given basis are not normalized.")
+        if not is_orthogonal(np.array([vector for vector in basis_vectors])):
```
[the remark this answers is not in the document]
Henri-ColibrITD: done

mpqp/core/instruction/measurement/basis.py
```diff
@@ -63,24 +86,42 @@ def __init__(
...
+            raise ValueError("The given basis is not orthogonal.")
```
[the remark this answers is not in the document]
Henri-ColibrITD: done

mpqp/core/instruction/measurement/basis.py
Henri-ColibrITD: agreed

mpqp/core/instruction/measurement/basis.py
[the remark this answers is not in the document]
Henri-ColibrITD: fixed for real

mpqp/tools/maths.py
```diff
@@ -74,6 +74,30 @@ def matrix_eq(lhs: Matrix, rhs: Matrix, atol: float = atol, rtol: float = rtol)
...
+        >>> is_orthogonal(np.array([[0,1],[1,0]]))
+        True
+        >>> is_orthogonal(np.array([[1,1],[1,0]]))
+        False
+        >>> is_orthogonal(rand_orthogonal_matrix(3))
+        True
+
+    """
+    return matrix_eq(
+        matrix.T.conj().dot(matrix), np.eye(len(matrix), dtype=np.complex64)
```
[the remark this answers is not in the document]
Henri-ColibrITD: this function was removed since it's only use was improper

mpqp/execution/runner.py
Henri-ColibrITD: hmm, good point, I think we are going to change the semantics of `job.circuit` because the alternative is extremely painful (and error prone)

mpqp/core/circuit.py
Henri-ColibrITD: should be fixed thanks to julien !

mpqp/core/instruction/measurement/basis.py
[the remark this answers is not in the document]
hJaffaliColibritd: Ok good

mpqp/core/circuit.py
hJaffaliColibritd: We wait for the merge of the solution from `mpqp-to-qasm`

mpqp/core/instruction/measurement/basis.py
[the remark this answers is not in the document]
hJaffaliColibritd: If we always check that before, then we good.

mpqp/execution/runner.py
hJaffaliColibritd: Done in `mpqp-to-qasm` branch, waiting for the merge in this branch.

mpqp/tools/maths.py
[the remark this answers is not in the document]
hJaffaliColibritd: Good

PR #102, opened 2024-10-22
Feat qasm 3 to 2

JulienCalistoTD:


mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +747,321 @@ def remove_include(qasm_code: str) -> str:
...
+        if m:
+            instructions_code += m.group(1) + ";\n"
+
+    return instructions_code, header_code
+
+
+@typechecked
+def open_qasm_3_to_2(
+    code: str,
+    included_tree_current_node: Node = Node("initial_code"),
```
Henri-ColibrITD, line 903: In python setting default params to objects is not recommended because this object would be the same (it would be the same reference to a mutable object) to all function call, so it you modify it, you would modify it for all future function call. Instead default it to `None` and in the body of the function, add a `if a is None: a = ...`
For an example of messed up default parameter initialization, you can for example run
```py
>>> def test(a = []): a.append(1); print(a)
>>> test()
[1]
>>> test()
[1, 1]
```

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +747,321 @@ def remove_include(qasm_code: str) -> str:
...
+    path_to_file: str = ".",
```
Henri-ColibrITD, line 904: same

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +747,321 @@ def remove_include(qasm_code: str) -> str:
...
+    defined_gates: set[str] = set(),
```
Henri-ColibrITD, line 905: same

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +747,321 @@ def remove_include(qasm_code: str) -> str:
...
     return replaced_code
+
+
+@typechecked
+def convert_instruction_3_to_2(
+    instr: str,
+    included_instr: set[Instr],
+    included_tree_current: Node,
+    defined_gates: set[str],
+    path_to_main: str = ".",
```
Henri-ColibrITD, line 758: same

tests/qasm/test_open_qasm_2_and_3.py
```diff
@@ -50,6 +51,37 @@ def test_circular_dependency_detection_false_positive():
...
+        )
+    assert "Circular dependency" in str(e.value)
+
+
+def test_circular_dependency_detection_false_positive_3_to_2():
+    try:
+        open_qasm_file_conversion_3_to_2(
+            qasm_folder + "circular_dep_a_3.qasm",
+        )
+    except RuntimeError:
```
Henri-ColibrITD, line 81: you sure you need to catch the error here ? If the error is not supposed to happen, I would just write
```py
def test_circular_dependency_detection_false_positive_3_to_2():
    open_qasm_file_conversion_3_to_2(qasm_folder + "circular_dep_a_3.qasm")
```

Henri-ColibrITD (changes requested):
see comments

Henri-ColibrITD:
are the tests passing ?

mpqp/qasm/open_qasm_2_and_3.py
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
JulienCalistoTD: done

tests/qasm/test_open_qasm_2_and_3.py
JulienCalistoTD: done

Henri-ColibrITD (approved):

hJaffaliColibritd:
When I try to execute this piece of code, 
```Python
>>>  c = """OPENQASM 3.0;
...: qubit[2] q;
...: gphase(0.3);
...: gate CX c, t {
...:    ctrl @ X c, t;
...: }
...: CX q[0], q[1];"""
>>>open_qasm_3_to_2(c)
```

I have this error `ValueError: Gates undefined at the time of usage: gphase, gphase`.

`gphase` is a built-in gate in OpenQasm3, and it looks like it is not handled in the code?

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +760,333 @@ def remove_include(qasm_code: str) -> str:
...
+                defined_gates,
+                path_to_main,
+            )
+            instructions_code += if_statement + " " + i_code
+            header_code += h_code
+    elif instr_name in {"reset", "barrier"}:
+        instructions_code += instr + ";\n"
+    else:
+        gate = instr.split()[0].split("(")[0]
+        if gate == "ctrl":
```
hJaffaliColibritd, line 906: The keyword `ctrl` is only known in OpenQASM2, so we cannot just put it like this I think, and I don't see direct way of overcoming this ... since creating new gates from existing by controlling is not available in OpenQASM2.
Do y'all have any idea ?

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +760,333 @@ def remove_include(qasm_code: str) -> str:
...
+            gate = instr.split()[2]
+        if gate not in defined_gates:
```
hJaffaliColibritd, line 908: Here, we will return an error for all the keyword/commands of OpenQASM3 that are not handled. For instance, if I have a for loop, or a switch case, or pulse and low level quantum operations, etc.
Should we put a todo here to treat later all those cases when we can (for loop can be unfolded, switch case to several if statements, ...).
Or we should ignore them and put a warning ?

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +760,333 @@ def remove_include(qasm_code: str) -> str:
...
+            )
+        m = re.match(r"\s*(.*)", instr)
+        if m:
+            instructions_code += m.group(1) + ";\n"
+
+    return instructions_code, header_code
+
+
+@typechecked
+def open_qasm_3_to_2(
```
hJaffaliColibritd, line 920: When we translate a circuit from OpenQASM3 to OpenQASM2, we lose the information of the global phase of the circuit. I think we should be able to estimate it through the conversions, and return it with the string (so it can be stored in a circuit, or post-processed by the user if needed).

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -31,9 +31,10 @@
 
 import os
```
hJaffaliColibritd, line 32: We should update the header and the sphinx doc to incorporate all the work done for 3 to 2.

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -46,21 +47,22 @@ class Instr(Enum):
...
+    RZZ = auto()
+    RXX = auto()
+    RCCX = auto()
+    RC3X = auto()
+    C3X = auto()
+    C4X = auto()
+    C3SQRTX = auto()
+    OQASM2_ALL_STDGATES = auto()
+    OQASM3_ALL_STDGATES = auto()
+    BRAKET_CUSTOM_INCLUDE = auto()
```
hJaffaliColibritd, line 65: AWS Braket has his own "funky" way of writing OpenQASM3 (`cx` is written `cnot` for instance, `phase` is written `phaseshift`, `ccx` is `ccnot`, `gphase` is not known, and many others). 

In the context of receiving an OpenQASM3 code from Braket circuit, we will have to handle that. We will have to detect when there are some keywords specific to braket and replace them by the right "original" OpenQASM3 one. Or we can also define a file that redefines all those keywords (`cnot`, ...) using native gates, like it is done in `braket_custom_include.inc`, but in the opposite way.

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +760,333 @@ def remove_include(qasm_code: str) -> str:
         else:
             replaced_code += line + "\n"
     return replaced_code
+
+
+@typechecked
+def convert_instruction_3_to_2(
```
hJaffaliColibritd, line 766: _I put here the comment because I didn't how where to put it and I wanted to be sure we can answer below._

When there is noise in the Braket circuit, and we would like to export that circuit to OpenQASM3, the noise models will be encoded in the OpenQASM3 using the `pragma` keyword (see example below).

I see two possible options to handle that : 
1. We grab the noise models directly from the Braket circuit before the conversion to OpenQASM3, to set it separately later in the QCircuit
2. We parse and analyse the pragmas to generate the right noisemodels in mpqp

In both cases, we can typically use a NoiseModel.from_other_language("str"(pragma) or BraketInstruction (which is a noise operator)).

To be discussed, if second choice, this will have impact on this PR.

hJaffaliColibritd (changes requested):
Good work in overall, not a easy task. But some changes needed, and some others comments need to be discussed.

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -290,7 +302,6 @@ def add_std_lib():
         )
         if new_instr is not None and new_instr not in included_instr:
             included_instr.add(new_instr)
-            header_code += qasm_code(new_instr)
```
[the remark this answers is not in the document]
JulienCalistoTD: we removed this line because we are adding new_instr to included_instr, which means that the code writes the same instruction twice, one time in included_instr and header_code

mpqp/qasm/open_qasm_2_and_3.py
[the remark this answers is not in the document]
JulienCalistoTD: done, without example

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +760,333 @@ def remove_include(qasm_code: str) -> str:
...
+    header_code = ""
+    instructions_code = ""
+
+    instr_match = re.match(r"\s*(\w+)\s*", instr)
+    if instr_match:
+        instr_name = instr_match.group(1)
+    else:
+        raise ValueError(f"Could not parse instruction: {instr}")
+
+    if instr.startswith("OPENQASM 2.0"):
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -108,6 +110,13 @@ class Instr(Enum):
     "phase",
     "sx",
 ]
+std_gates_3_to_2_map = {
+    "u3": "u",
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -108,6 +110,13 @@ class Instr(Enum):
...
+    "cp": "cu1",
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +760,333 @@ def remove_include(qasm_code: str) -> str:
...
+        m = re.match(
+            r"\s*([\w\d_]+)(\[.*?\])?\s*=\s*measure\s*([\w\d_]+)(\[.*?\])?\s*", instr
+        )
+        if m:
+            c, nb_c, q, nb_q = m.groups()
+            if nb_c and nb_q:
+                instructions_code += f"measure {q}{nb_q} -> {c}{nb_c};\n"
+            else:
+                instructions_code += f"measure {q} -> {c};\n"
+    elif instr_name == "u3":
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +760,333 @@ def remove_include(qasm_code: str) -> str:
...
+        instructions_code += re.sub(r"\bu3\b", "u", instr) + ";\n"
+    elif instr_name == "cp":
+        header_code += add_qe_lib()
+        instructions_code += re.sub(r"\bcp\b", "cu1", instr) + ";\n"
+    elif instr_name in std_gates_3_to_2_map:
+        converted_instr_name = std_gates_3_to_2_map[instr_name]
+        instructions_code += (
+            re.sub(r"\b" + instr_name + r"\b", converted_instr_name, instr) + ";\n"
+        )
+    elif instr_name in std_gates_3:
```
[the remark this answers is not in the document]
JulienCalistoTD: removed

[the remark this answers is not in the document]
JulienCalistoTD: removed

mpqp/qasm/open_qasm_2_and_3.py
JulienCalistoTD: removed

mpqp/qasm/open_qasm_2_and_3.py
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
```diff
@@ -740,3 +760,333 @@ def remove_include(qasm_code: str) -> str:
...
+def open_qasm_3_to_2(
+    code: str,
+    included_tree_current_node: Optional[Node] = None,
+    path_to_file: Optional[str] = None,
+    defined_gates: Optional[set[str]] = None,
+) -> str:
+    """Converts an OpenQASM 3.0 code back to OpenQASM 2.0.
+
+    This function will also recursively go through the imported files to
+    translate them too. It is a partial conversion (the ``opaque`` keyword is
```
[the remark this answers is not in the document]
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
JulienCalistoTD: done

mpqp/qasm/open_qasm_2_and_3.py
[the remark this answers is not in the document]
hJaffaliColibritd: Ok makes sense, thanks

mpqp/qasm/open_qasm_2_and_3.py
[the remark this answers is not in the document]
hJaffaliColibritd: Example added

PR #106, opened 2024-11-18
test: validity for gate, job_type, measure, noise, pauli_string and other_instructions 

JulienCalistoTD:


mpqp/tools/circuit.py
```diff
@@ -159,6 +171,53 @@ def random_instruction(
...
+    """
+    rng = np.random.default_rng(seed)
+
+    if noise_model is None:
+        noise_model = NOISE_MODELS
+
+    noise = rng.choice(np.array(noise_model))
+
+    if issubclass(noise, AmplitudeDamping):
+        prob = rng.uniform(0, 1)
```
Henri-ColibrITD, line 203: I'm not in love with this, I guess this is the reason why you asked if we couldn't have a `proba` attribute for all noises

tests/execution/test_validity.py
```diff
@@ -292,3 +323,163 @@ def test_observable_ideal_case(
...
+        elif language in [Language.MY_QLM] and not isinstance(
+            noise_build, (Depolarizing, PhaseDamping)
+        ):
+            with pytest.raises(NotImplementedError):
+                noise_build.to_other_language(language)
+        else:
+            assert noise_build.to_other_language(language) is not None
+
+
+other_instr = [
```
Henri-ColibrITD, line 468: I would have that as a fixture

tests/execution/test_validity.py
```diff
@@ -292,3 +323,163 @@ def test_observable_ideal_case(
...
+            )  # pyright: ignore[reportUnnecessaryIsInstance]
+            and not isinstance(measure.basis, ComputationalBasis)
+        ):
+            with pytest.raises(NotImplementedError):
+                measure.to_other_language(language)
+        else:
+            assert measure.to_other_language(language) is not None
+
+
+pauli_strings = [Ip @ Xp @ Yp @ Zp, Xp + Zp, Yp]
```
Henri-ColibrITD, line 442: fixture

Henri-ColibrITD (changes requested):
tiny tweaks

tests/execution/test_validity.py
JulienCalistoTD: done

tests/execution/test_validity.py
JulienCalistoTD: done

PR #115, opened 2024-11-22
Fix qiskit update

Henri-ColibrITD:
since it's a patch, I'm suggesting to merge this directly on main


setup.py
```diff
@@ -33,6 +33,7 @@
     entry_points={
         "console_scripts": [
             "setup_connections = mpqp_scripts.setup_connections:main_setup",
+            "qiskit_update = mpqp_scripts.qiskit_update:update_packages",
```
hJaffaliColibritd, line 36:
```suggestion
            "update_qiskit = mpqp_scripts.qiskit_update:update_packages",
```

mpqp_scripts/update_qiskit.py
```diff
@@ -0,0 +1,42 @@
+import subprocess
+import sys
+
+qiskit_packages = [
```
hJaffaliColibritd, line 4: _Bonus remark, we can address it later._
Do you think we can build this list dynamically for each user ? In case of our list is not exhaustive.
Something like ``pip list | grep qiskit`` and ``pip list | grep ibm``.

mpqp/execution/providers/ibm.py
```diff
@@ -21,7 +21,6 @@
 from mpqp.execution.devices import AZUREDevice, IBMDevice
 from mpqp.execution.job import Job, JobStatus, JobType
 from mpqp.execution.result import Result, Sample, StateVector
-from mpqp.execution.simulated_devices import IBMSimulatedDevice
```
hJaffaliColibritd, line 24: Question: Why did we remove the import of this mpqp class to deal with it as it was a class from an external package? Did cause any circular import problem? Or maybe for performance issues ? (the generation of fake devices?)

hJaffaliColibritd (changes requested):

mpqp_scripts/update_qiskit.py
Henri-ColibrITD: I'd say we put this on trello

mpqp/execution/providers/ibm.py
Henri-ColibrITD: It's a performance issue: in order to generate the enum values, you have to import qiskit, and we don't want that to happen on default import (import mpqp)

mpqp_scripts/update_qiskit.py
hJaffaliColibritd: Agreed, done. I resolve

mpqp/execution/providers/ibm.py
hJaffaliColibritd: Very good. I resolve

hJaffaliColibritd (approved):

PR #116, opened 2024-11-22
Fix typecheck errors 

MoHermes:
Resolve type-checking errors encountered during test execution.

### DIFF
--- a/mpqp/tools/circuit.py
+++ b/mpqp/tools/circuit.py
@@ -39,8 +39,7 @@ if TYPE_CHECKING:
     from qiskit.circuit.quantumcircuitdata import CircuitInstruction
 
 
-# @typechecked
-# FIXME: Resolve type-checking errors encountered during test execution.
+@typechecked
 def random_circuit(
     gate_classes: Optional[Sequence[type[Gate]]] = None,
     nb_qubits: int = 5,
@@ -85,7 +84,7 @@ def random_circuit(
     rng = np.random.default_rng(seed)
 
     if nb_gates is None:
-        nb_gates = rng.integers(5, 10)
+        nb_gates = int(rng.integers(5, 10))
 
     qcircuit = QCircuit(nb_qubits)
     for _ in range(nb_gates):
--- a/mpqp/tools/circuit.py
+++ b/mpqp/tools/circuit.py
@@ -84,7 +84,9 @@ def random_circuit(
     rng = np.random.default_rng(seed)
 
     if nb_gates is None:
-        nb_gates = int(rng.integers(5, 10))
+        nb_gates = rng.integers(5, 10)
+        if TYPE_CHECKING:
+            assert isinstance(nb_gates,int)
 
     qcircuit = QCircuit(nb_qubits)
     for _ in range(nb_gates):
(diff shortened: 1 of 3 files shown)


mpqp/tools/circuit.py
```diff
@@ -85,7 +84,7 @@ def random_circuit(
     rng = np.random.default_rng(seed)
 
     if nb_gates is None:
-        nb_gates = rng.integers(5, 10)
+        nb_gates = int(rng.integers(5, 10))
```
hJaffaliColibritd, line 87: better use 
```Python
if TYPE_CHECKING:
    assert isinstance(nb_gates,int)
```

mpqp/tools/display.py
```diff
@@ -57,18 +56,20 @@ def with_sign(val: np.complex64) -> str:
     return "+ " + str_rounded
 
 
-# @typechecked
-# FIXME: Resolve type-checking errors encountered during test execution.
-def _remove_null_imag(val: np.complex64) -> np.complex64 | np.float32 | int:
+@typechecked
+def _remove_null_imag(
+    val: np.complex64 | np.complex128,
+) -> np.complex64 | np.complex128 | np.float32 | int:
```
hJaffaliColibritd, line 62: What if we create a generic type for all those complex/float/int ?

mpqp/tools/theoretical_simulation.py
```diff
@@ -233,7 +232,7 @@ def validate_noisy_circuit(
     Returns:
         Weather our noise pipeline matches the theory or not.
     """
-    return exp_id_dist(circuit, shots, device) <= trust_int(circuit)
+    return bool(exp_id_dist(circuit, shots, device) <= trust_int(circuit))
```
hJaffaliColibritd, line 235: I would use also `if TYPE_CHECKING`.

hJaffaliColibritd (changes requested):

mpqp/tools/circuit.py
hJaffaliColibritd:
```suggestion
        nb_gates = rng.integers(5, 10)
        if TYPE_CHECKING:
            assert isinstance(nb_gates,int)
```

mpqp/tools/theoretical_simulation.py
Henri-ColibrITD: why is the bool needed ?

MoHermes: Because of this error:
typeguard.TypeCheckError: the return value (numpy.bool_) is not an instance of bool

mpqp/tools/circuit.py
MoHermes: I already done that before cast int, but really don't work! that's why

MoHermes: I tested with the updated you done @hJaffaliColibritd, still not passing
image

hJaffaliColibritd: You right, we put it back to int and that's it

hJaffaliColibritd: Oh, it seems that it returns a numpy.int65

mpqp/tools/theoretical_simulation.py
hJaffaliColibritd: Ok for me

hJaffaliColibritd (approved):

PR #119, opened 2024-12-05
feat: to_other_language cirq without QASM2.0

JulienCalistoTD:


mpqp/core/circuit.py
```diff
@@ -1134,12 +1134,46 @@ def to_other_language(
...
-                assert isinstance(qasm2_code, str)
-            from mpqp.qasm.qasm_to_cirq import qasm2_to_cirq_Circuit
+            from cirq.circuits.circuit import Circuit as CirqCircuit
+            from cirq.ops.named_qubit import NamedQubit
+            from cirq.ops.identity import I
+
+            cirq_qubits = [NamedQubit(f"q_{i}") for i in range(self.nb_qubits)]
+            cirq_circuit = CirqCircuit()
+
+            for qubit in cirq_qubits:
```
hJaffaliColibritd, line 1144: I would do this only add the end of the instantiation of the circuit, so we do it only for the qubits that were not used (like in braket). To avoid any useless identity.

mpqp/core/circuit.py
```diff
@@ -1134,12 +1134,46 @@ def to_other_language(
...
+                elif isinstance(instruction, CustomGate):
+                    custom_circuit = QCircuit(self.nb_qubits)
+                    custom_circuit.add(instruction)
+                    qasm2_code = custom_circuit.to_other_language(Language.QASM2)
+                    if TYPE_CHECKING:
+                        assert isinstance(qasm2_code, str)
+                    from mpqp.qasm.qasm_to_cirq import qasm2_to_cirq_Circuit
+
+                    custom_cirq_circuit = qasm2_to_cirq_Circuit(qasm2_code)
+                    cirq_circuit += custom_cirq_circuit
```
hJaffaliColibritd, line 1159: We are sure that the concatenation operation works well in for Cirq circuit ?

mpqp/core/circuit.py
```diff
@@ -1134,12 +1134,46 @@ def to_other_language(
...
+                elif isinstance(instruction, ControlledGate):
+                    targets = []
+                    for target in instruction.targets:
+                        targets.append(cirq_qubits[target])
+                    controls = []
+                    for control in instruction.controls:
+                        controls.append(cirq_qubits[control])
+                    cirq_instruction = instruction.to_other_language(Language.CIRQ)
+                    cirq_circuit.append(cirq_instruction.on(*controls, *targets))
+                else:
```
hJaffaliColibritd, line 1170: Here we handle all the other cases ? (Parametrized gate, Barrier, BasisMeasure, ..) right ?

hJaffaliColibritd (commented):
Only reviewed `circuit.py`

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -182,6 +183,14 @@ def braket_gate(
     ]:
         pass
 
+    @classproperty
```
assombrance, line 186: wasn't it a problem to be both of those ?

Henri-ColibrITD, line 186: isn't it a problem for the typechecker to put these two together ?

mpqp/execution/providers/google.py
```diff
@@ -298,7 +298,7 @@ def extract_result_SAMPLE(
 
     data = [
         Sample(
-            bin_str="".join(map(str, state)),
+            bin_str="".join(map(bin, state)),
```
Henri-ColibrITD, line 301: bin adds "0b" at the beginning, is that correct ?

mpqp/core/circuit.py
JulienCalistoTD: The concatenation operation in Cirq works with the names of the qubits, and it functions correctly if we have the right name for each qubit. We need all the qubits, so we must apply the identity gate to each qubit.

mpqp/core/circuit.py
JulienCalistoTD: we need all the qubits for concatenation operation, so we must keep it like this

mpqp/core/circuit.py
JulienCalistoTD: Cirq doesn't have a Barrier operation, so we skips it. breakpoints is also skip. While the rest is handled properly.

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: No, it isn't a problem

mpqp/execution/providers/google.py
JulienCalistoTD: yes, in Sample we do :  `int('0b100', 2)` so it works

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -640,6 +691,15 @@ def qiskit_gate(cls):
...
         return CPhaseGate
 
+    @classproperty
+    def cirq_gate(cls):
+        from cirq.ops.common_gates import ZPowGate
+        from cirq.ops.controlled_gate import ControlledGate as CirqControlledGate
+
+        return lambda theta: CirqControlledGate(ZPowGate(exponent=theta / np.pi))
+
+        return controlled_phase
```
hJaffaliColibritd, line 701: This line is useless right ?

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -868,6 +946,40 @@ def qiskit_gate(cls):
...
         return UGate
 
+    @classproperty
+    def cirq_gate(cls):
+        from cirq.circuits.qasm_output import QasmUGate
+        from cirq.ops.common_gates import ry as cirq_ry, rz as cirq_rz
+        from cirq.ops.global_phase_op import GlobalPhaseGate
+        from cirq.ops.raw_types import Qid
+
+        class CirqUGate(QasmUGate):  # pyright: ignore[reportUntypedBaseClass]
```
hJaffaliColibritd, line 956: The class will be redifined each time we call the property. Isn't it better if we define the class outside and only return it when needed ?

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -868,6 +946,40 @@ def qiskit_gate(cls):
...
+                self, theta, phi, lmda  # pyright: ignore[reportMissingParameterType]
+            ) -> None:
+                self.lmda = lmda
+                self.theta = theta
+                self.phi = phi
+
+            def __repr__(self) -> str:
+                return (
+                    f'U('
+                    f'theta={self.theta !r}, '
```
hJaffaliColibritd, line 967: Why `!r` is used for theta and hi, and not for lambda ?

Another question concerning those angles, no need to tweak them with some division by pi ? When you can ry and rz in the decompose ?

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -1117,6 +1249,12 @@ def qiskit_gate(cls):
 
         return PhaseGate
 
+    @classproperty
+    def cirq_gate(cls):
+        from cirq.ops.common_gates import ZPowGate
+
+        return lambda theta: ZPowGate(exponent=theta / np.pi)
```
hJaffaliColibritd, line 1256: How do you make the link between `k` and `theta` here ?

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -1199,6 +1337,12 @@ def qiskit_gate(cls):
...
```
hJaffaliColibritd, line 1344: Same question as for `Rk`

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -1385,6 +1541,13 @@ def qiskit_gate(cls):
 
         return CPhaseGate
 
+    @classproperty
+    def cirq_gate(cls):
+        from cirq.ops.controlled_gate import ControlledGate as CirqControlledGate
+        from cirq.ops.common_gates import ZPowGate
+
+        return lambda theta: CirqControlledGate(ZPowGate(exponent=theta / np.pi))
```
hJaffaliColibritd, line 1549: Same question as for `Rk`

mpqp/tools/maths.py
```diff
@@ -161,7 +161,7 @@ def closest_unitary(matrix: Matrix):
 
 
 @typechecked
-def cos(angle: Expr | Real) -> sp.Expr | float:
+def cos(angle: Expr | Real | float) -> sp.Expr | float:
```
hJaffaliColibritd, line 164: In principle, no need to add `float` here, as it is "contained" in `Real`

mpqp/tools/maths.py
```diff
@@ -186,7 +186,7 @@ def cos(angle: Expr | Real) -> sp.Expr | float:
 
 
 @typechecked
-def sin(angle: Expr | Real) -> sp.Expr | float:
+def sin(angle: Expr | Real | float) -> sp.Expr | float:
```
hJaffaliColibritd, line 189: same as above

mpqp/tools/maths.py
```diff
@@ -211,7 +211,7 @@ def sin(angle: Expr | Real) -> sp.Expr | float:
 
 
 @typechecked
-def exp(angle: Expr | Complex) -> sp.Expr | complex:
+def exp(angle: Expr | Complex | complex) -> sp.Expr | complex:
```
hJaffaliColibritd, line 214: Same remark for `complex` and `Complex`

hJaffaliColibritd (changes requested):

mpqp/core/circuit.py
JulienCalistoTD: The BasisMeasure is a simple measurement with a pre-measurement circuit, we juste have to concatenate it. so yes

mpqp/tools/maths.py
JulienCalistoTD: yes, but i had typcheck issues

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: it's a lamba function we pass theta in the RotationGate class to_otherLanguages()

Henri-ColibrITD (approved):

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: done

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: because it is an end of line so does not need to be added at the end.

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: because it is an end of line so does not need to be added at the end.

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: Same answer as for Rk

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: Same answer as for Rk

mpqp/tools/maths.py
JulienCalistoTD: same as above

mpqp/tools/maths.py
JulienCalistoTD: same as above

mpqp/execution/providers/google.py
hJaffaliColibritd: Is this to be resolved @Henri-ColibrITD  ?

mpqp/tools/maths.py
hJaffaliColibritd: so at the end Real from the package number worked no ? This join the same discussion as Complex from numbers

mpqp/tools/maths.py
hJaffaliColibritd: this worked at the end right ?

hJaffaliColibritd (changes requested):

mpqp/core/circuit.py
hJaffaliColibritd: We said we introduce a boolean argument in the function so the user can decice or not to add the pre-measurement, and by default it is set to True to respect the semantics of the circuit

JulienCalistoTD: Done: Pre-measurement is added by default during `to_other_language` with the variable `skip_pre_measure: bool = False` instead of `generate_job`

mpqp/tools/maths.py
JulienCalistoTD: Remove `Real` and remplace by `float`

mpqp/tools/maths.py
JulienCalistoTD: Remove `Real` and remplace by `float`

mpqp/tools/maths.py
JulienCalistoTD: Remove `Complex` and remplace by `complex`

mpqp/tools/maths.py
hJaffaliColibritd: But if a user tries `cos(1)` instead of `cos(1.0)`, this will raise an Error

mpqp/tools/maths.py
hJaffaliColibritd: What if a user put a float or int angle ? This will raise an error. I would let Complex

mpqp/tools/maths.py
JulienCalistoTD: no, there's no Error

mpqp/tools/maths.py
JulienCalistoTD: for float, int or complex , there's no Error

mpqp/core/circuit.py
```diff
@@ -345,23 +355,33 @@ def _update_targets_components(self, component: Instruction | NoiseModel):
...
+
     @nb_qubits.setter
     def nb_qubits(self, nb_qubits: int):
         if self._user_nb_qubits is None or self._user_nb_qubits != nb_qubits:
             self._user_nb_qubits = nb_qubits
             self._set_nb_qubits_dynamic(nb_qubits)
 
+    @nb_cbits.setter
+    def nb_cbits(self, nb_cbits: int):
+        if self._user_nb_cbits is None or self._user_nb_cbits != nb_cbits:
```
hJaffaliColibritd, line 382: If `nb_cbits < _user_nb_cbits` we should return an error right ?

mpqp/core/circuit.py
```diff
@@ -956,11 +976,26 @@ def without_noises(self) -> QCircuit:
...
+        return circuit
+
     def to_other_language(
         self,
         language: Language = Language.QISKIT,
         cirq_proc_id: Optional[str] = None,
         translation_warning: bool = True,
+        skip_pre_measure: bool = False,
     ) -> QuantumCircuit | myQLM_Circuit | braket_Circuit | cirq_Circuit | str:
         """Transforms this circuit into the corresponding circuit in the language
```
hJaffaliColibritd, line 1000: We should add in the docstring that by default the pre-measure is included, and also documen the new argument `skip_pre_measure`, and give example also

mpqp/execution/providers/google.py
```diff
@@ -136,7 +136,7 @@ def run_local(job: Job) -> Result:
         return run_local_processor(job)
 
     if job.job_type == JobType.STATE_VECTOR:
-        circuit = job.circuit.without_measurements()
+        circuit = job.circuit.without_measurements() + job.circuit.pre_measure()
         cirq_circuit = circuit.to_other_language(Language.CIRQ)
```
hJaffaliColibritd, line 140: Same remark as above, we should skip the pre_measure here if we put it manually above, no ?

mpqp/execution/providers/ibm.py
```diff
@@ -438,7 +438,9 @@ def run_aer(job: Job):
         backend_sim = AerSimulator(method=job.device.value)
 
     qiskit_circuit = (
-        job_circuit.without_measurements().to_other_language(Language.QISKIT)
+        (
+            job_circuit.without_measurements() + job_circuit.pre_measure()
+        ).to_other_language(Language.QISKIT)
         if (job.job_type == JobType.STATE_VECTOR)
         else job_circuit.to_other_language(Language.QISKIT)
```
hJaffaliColibritd, line 445: Same remark as above, we should skip the pre_measure here if we put it manually above, no ?

mpqp/tools/maths.py
Henri-ColibrITD: > But if a user tries `cos(1)` instead of `cos(1.0)`, this will raise an Error right ? I would let Real instead

Here Julien is adding an option, not replacing the existing one, so I don't see any reason new error would arise. It's sad we have to do this but I think we already encountered this kind of error. Can you tell us where you encountered the error Julien ? so we check if there is no other solution ?

mpqp/tools/maths.py
Henri-ColibrITD: Why do you have to replace it ? just adding the option isn't enough ?

mpqp/tools/maths.py
Henri-ColibrITD: maybe having this conversation only once is enough x)
Raising the concern at each occurrence can be useful but then no need to discuss the same thing multiple times

mpqp/tools/maths.py
Henri-ColibrITD: Ah, no, sorry, I didn't see it was indeed removed. I just checked and it worked. I don't know why it didn't work before, weird...

Henri-ColibrITD: this is OK for me

mpqp/core/circuit.py
```diff
@@ -130,14 +130,15 @@ def __init__(
...
             data = []
-        self.nb_cbits = nb_cbits
-        """See parameter description."""
         self.label = label
         """See parameter description."""
         self.instructions: list[Instruction] = []
         """List of instructions of the circuit."""
         self.noises: list[NoiseModel] = []
         """List of noise models attached to the circuit."""
+        self._user_nb_cbits: Optional[int] = None
```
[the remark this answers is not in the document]
Henri-ColibrITD: nop, cbits vs qbits

mpqp/execution/providers/azure.py
```diff
@@ -38,7 +38,9 @@ def run_azure(job: Job) -> Result:
     from qiskit import QuantumCircuit
 
     qiskit_circuit = (
-        job.circuit.without_measurements().to_other_language(Language.QISKIT)
+        (
+            job.circuit.without_measurements() + job.circuit.pre_measure()
+        ).to_other_language(Language.QISKIT)
```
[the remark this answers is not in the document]
Henri-ColibrITD: yes, I think you're right, I'll let Julien confirm though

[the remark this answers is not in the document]
JulienCalistoTD: no because we remove measurements with `without_measurements()`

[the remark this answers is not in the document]
Henri-ColibrITD: oh, right !

mpqp/core/circuit.py
```diff
@@ -273,6 +276,15 @@ def _check_components_targets(self, components: Instruction | NoiseModel):
                     "In noisy circuits, BasisMeasure must span all qubits in the circuit."
                 )
 
+    def _update_cbits(self, cbits: int):
```
Henri-ColibrITD, line 279: hmmm, I'd like to discuss the logic of this function with you Julien

mpqp/core/circuit.py
Henri-ColibrITD: ok for me (error message cleaned in commit 84bef)

mpqp/core/circuit.py
Henri-ColibrITD: `>` is weird to me

PR #120, opened 2024-12-17
feat: dynamic circuit, and fix VariableSizeBasis

JulienCalistoTD:


mpqp/core/circuit.py
```diff
@@ -210,15 +212,23 @@ def add(self, components: OneOrMany[Instruction | NoiseModel]):
...
-            component_type = (
-                "Instruction" if isinstance(components, Instruction) else "Noise model"
-            )
-            raise NumberQubitsError(
-                f"{component_type} {type(components)}'s connections "
-                f"({components.connections()}) are not compatible with circuit"
-                f" size ({self.nb_qubits})."
-            )
+        if self._user_nb_qubits is not None:
+            if any(conn >= self.nb_qubits for conn in components.connections()):
```
hJaffaliColibritd, line 216: Since `self._user_nb_qubits is not None`, you can directly use `self._user_nb_qubits` instead of calling the property `self.nb_qubits`

mpqp/core/circuit.py
```diff
@@ -1280,12 +1297,15 @@ def __str__(self) -> str:
 
     def __repr__(self) -> str:
         instructions_repr = ", ".join(repr(instr) for instr in self.instructions)
+        nb_qubits = ""
+        if self._user_nb_qubits is not None:
```
hJaffaliColibritd, line 1301: If you call the property `self.nb_qubits`, then this check is already done there, so you can directly use the property for the repr no ?

mpqp/core/instruction/measurement/basis.py
```diff
@@ -165,24 +165,50 @@ def to_computational(self):
...
+
+    Example:
+        >>> custom_basis = VariableSizeBasis([np.array([1,0]), np.array([0,-1])], symbols=("↑", "↓"))
+        >>> custom_basis.pretty_print()
+        Basis: [
+            [1, 0],
+            [0, -1]
+        ]
+        >>> circ = QCircuit([X(0), H(1), CNOT(1, 2), Y(2)])
+        >>> circ.add(BasisMeasure([0, 1, 2], basis=custom_basis, shots=10000))
```
hJaffaliColibritd, line 185: We can also remove the targets here in `BasisMeasure`

mpqp/core/instruction/measurement/basis.py
```diff
@@ -195,10 +221,12 @@ def set_size(self, nb_qubits: int):
         Args:
             nb_qubits: number of qubits in the basis
         """
-        pass
+        Bn = reduce(np.kron, [self.basis_vectors] * nb_qubits, np.eye(1))
```
hJaffaliColibritd, line 224: chipotage: maybe add a condition that, when the number of qubits is already matching the size of basis vector, we don't need to perform all the steps below (typically when we call `set_size` in the __init__)

mpqp/tools/display.py
```diff
@@ -143,7 +143,10 @@ def format_element(element: Union[int, float, complex | Expr], round: int = 5) -
 
 @typechecked
 def clean_1D_array(
-    array: list[complex] | npt.NDArray[np.complex64 | np.float32], round: int = 5
+    array: (
+        list[complex | float | int] | npt.NDArray[np.complex64 | np.float32 | np.int32]
```
hJaffaliColibritd, line 147: Maybe instead of putting `complex | float | int`, we can use the class `Complex` from the package `numbers`?

hJaffaliColibritd (commented):
Good and important work.
Should we add more tests for Basis/VariableSizeBasis ?

mpqp/core/circuit.py
```diff
@@ -344,10 +354,14 @@ def _update_targets_components(self, component: Instruction | NoiseModel):
...
-        return self._nb_qubits
+        return self._nb_qubits if self._user_nb_qubits is None else self._user_nb_qubits
 
     @nb_qubits.setter
     def nb_qubits(self, nb_qubits: int):
+        self._user_nb_qubits = nb_qubits
+        self._set_nb_qubits_dynamic(nb_qubits)
+
+    def _set_nb_qubits_dynamic(self, nb_qubits: int):
         self._nb_qubits = nb_qubits
```
Henri-ColibrITD, line 365: maybe for efficiency sake we can check if it was not already the same before, and if so we do nothing ?

mpqp/core/circuit.py
```diff
@@ -1280,12 +1297,15 @@ def __str__(self) -> str:
...
     def __repr__(self) -> str:
         instructions_repr = ", ".join(repr(instr) for instr in self.instructions)
+        nb_qubits = ""
+        if self._user_nb_qubits is not None:
+            nb_qubits = f", nb_qubits={self.nb_qubits}"
 
         if self.noises:
             noise_repr = ", ".join(map(repr, self.noises))
-            return f'QCircuit([{instructions_repr}, {noise_repr}], nb_qubits={self.nb_qubits}, nb_cbits={self.nb_cbits}, label="{self.label}")'
+            return f'QCircuit([{instructions_repr}, {noise_repr}]{nb_qubits}, nb_cbits={self.nb_cbits}, label="{self.label}")'
```
Henri-ColibrITD, line 1306: if no noise we would get an comma trailing list here, and if no instruction too we would get the invalid syntax `[,]` I believe

Henri-ColibrITD: we should only display the comma between intrs and noises if noises are not empty

mpqp/core/circuit.py
```diff
@@ -1280,12 +1297,15 @@ def __str__(self) -> str:
 
     def __repr__(self) -> str:
         instructions_repr = ", ".join(repr(instr) for instr in self.instructions)
+        nb_qubits = ""
+        if self._user_nb_qubits is not None:
+            nb_qubits = f", nb_qubits={self.nb_qubits}"
 
         if self.noises:
```
Henri-ColibrITD, line 1304: I would to something like 
```py
components = instructions_repr + ("" if self.noises is None else (", " + ", ".join(map(repr, self.noises))))
return f'QCircuit([components]{nb_qubits}, nb_cbits={self.nb_cbits}, label="{self.label}")'
```

mpqp/core/instruction/measurement/basis.py
```diff
@@ -165,24 +165,50 @@ def to_computational(self):
...
     """A variable-size basis with a dynamically adjustable size to different qubit numbers
     during circuit execution.
 
     Args:
         nb_qubits: number of qubits in the basis. If not provided,
             the basis can be dynamically sized later using the `set_size` method.
-
         symbols: custom symbols for representing basis states, defaults to ("0", "1").
+
+    Example:
```
Henri-ColibrITD, line 177: very nice

mpqp/tools/display.py
Henri-ColibrITD: In theory I agree, in practice I had many problems with the abstract number types, but it would be cool if it worked (to try anyway ?)

Henri-ColibrITD (changes requested):
see comments

mpqp/core/circuit.py
JulienCalistoTD: changed _user_nb_qubits to nb_qubits

mpqp/core/circuit.py
Henri-ColibrITD: was this taken into account ?

mpqp/tools/display.py
Henri-ColibrITD: have you checked it indeed works in many different contexts ?

Henri-ColibrITD: haaaa you defined your own complex type, I didn't notice, have you discussed this choice with Hamza ? I don't know how to feel about it (on one hand, it makes the code more readable here, but since it's not used anywhere else, maybe an overcomplication ? and on the other hand, since it's the same name as a "standard class" -- `numbers.Complex`, it may be confusing)

mpqp/core/circuit.py
Henri-ColibrITD: taking this into account ?

tests/core/instruction/measurement/test_basis.py
Henri-ColibrITD, line 1: I didn't see a reply on this comment:
https://github.com/ColibrITD-SAS/mpqp/pull/120#pullrequestreview-2525114520

Henri-ColibrITD (commented):
it would be good to reply to all comments, but we're very close to merge now

mpqp/core/circuit.py
JulienCalistoTD:
```py
    @property
    def nb_qubits(self) -> int:
        """Number of qubits of the circuit."""
        return self._nb_qubits if self._user_nb_qubits is None else self._user_nb_qubits
```
Yes, self.nb_qubits will always return a value even if self._user_nb_qubits is None, but that's not what we wanted to check. We want to know if the circuit is dynamic, so if self._user_nb_qubits is None

mpqp/tools/display.py
JulienCalistoTD: Finally it works with the `Complex` class from the `numbers` package, so I removed my class. This is better

mpqp/core/circuit.py
JulienCalistoTD: add check : 
```py
    if self._user_nb_qubits is None or self._user_nb_qubits != nb_qubits:
```

tests/core/instruction/measurement/test_basis.py
JulienCalistoTD: i add some test for VariableSizeBasis and allows basis_vectors of any size and set_size only on multiple of basis_vectors.

mpqp/core/circuit.py
hJaffaliColibritd: I still don't get it. Indeed here you check if the circuit is dynamic, but inside the if, you only call self.nb_qubits, that will check again if the circuit is dynamic and return the right value.

What you mean is that you want nb_qubits to stay the empty string if the circuit is dynamic, and otherwise print the self._user_nb_qubits?

JulienCalistoTD: Yes, if the circuit is dynamic, we want to create a dynamic circuit from repr, so we must not specify a number of qubits, nb_circuit should not be initialized.

hJaffaliColibritd: Ok thanks

hJaffaliColibritd (approved):

PR #122, opened 2025-02-12
Fix pauli monomial symbol

JulienCalistoTD:


mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -210,6 +211,40 @@ def __eq__(self, other: object) -> bool:
 
         return self.to_dict() == other.to_dict()
 
+    def subs(
+        self, values: dict[Expr | str, Complex], remove_symbolic: bool = True
```
hJaffaliColibritd, line 215: I would put `Real` instead of `Complex`, because we always supposed that we implemented the `PauliString` for representing observables, so the coefficients can't be `complex`.

In reality, the values of the symbolic variables can be complex, but their combination, and the coefficient of the monomials after substitution should be real. Otherwise, we force to take the real part each time.

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -32,7 +34,10 @@
     from qat.core.wrappers.observable import Term
     from qiskit.quantum_info import SparsePauliOp
 
+FixedReal = Union[Real, float, Expr]
```
hJaffaliColibritd, line 37: I am not sure we should include `Expr` here, is there any particular reason?

JulienCalistoTD: I'm not sure I understood that comment well. The main purpose of the PR is to add a symbol and a symbol comes with `Expr`, so that was the main purpose of adding it here.

mpqp/core/instruction/measurement/pauli_string.py
JulienCalistoTD: done: replace `Complex` byr `Real` in e85113b0fc434e034f23ddfcc2897596e053e62e

mpqp/core/instruction/measurement/pauli_string.py
JulienCalistoTD: ok i can put `Coef = Union[FixedReal, Expr]` for exemple but i need to change every `FixedReal` to ` Coef`. So we will not use `FixedReal`

hJaffaliColibritd: Ummm, indeed if `FixedReal` is not used elsewhere, then maybe we keep `Union[Real, float, Expr]` and just rename it `Coef` using the refactor feature of your IDE ?

JulienCalistoTD: done: rename in 58b8e2e884410b3d731f4003eeff00f5f4b70ba0

Henri-ColibrITD (approved):

PR #126, opened 2025-03-11
Feat multi-observable for IBM and diagonal observables

MoHermes:


hJaffaliColibritd (approved):

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -66,66 +71,187 @@ class Observable:
 
     """
 
-    def __init__(self, observable: Matrix | PauliString):
+    def __init__(self, observable: Matrix | PauliString | list[Real]):
```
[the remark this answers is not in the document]
MoHermes: Here a232e59

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -66,66 +71,187 @@ class Observable:
...
+            # We first check if the pauli string is already known, we use it for efficiency
+            if self._pauli_string is not None:
+                self._is_diagonal = self._pauli_string.is_diagonal()
+            # If not we check if the matrix is already known,
+            elif self._matrix is not None:
+                self._is_diagonal = is_diagonal(self._matrix)
+            # If only the diagonal elements are known, it means the observable is diagonal
+            elif self._diag_elements is not None:
+                self._is_diagonal = True
+            # Otherwise, the observable is empty, we return False by convention
```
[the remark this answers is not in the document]
MoHermes: It really depends on how we want to handle an empty observable. Right now, the function simply returns `False` if none of `_pauli_string`, `_matrix`, or `_diag_elements` are set,
If an observable being empty is an unexpected case, then raisin an error would be better!!

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -796,6 +811,23 @@ def __hash__(self):
...
+
+        Examples:
+            >>> from mpqp.measures import I, X, Y, Z
+            >>> (I @ X + Z @ Y - Y @ X).is_diagonal()
+            False
+            >>> (I @ Z @ I - 2* Z @ Z @ Z + I @ I @ I).is_diagonal()
+            True
+
+        """
+        return all([all([a != X and a != Y for a in m.atoms]) for m in self.monomials])
```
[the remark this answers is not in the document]
MoHermes: Yeah you are right 27f9b24

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -66,66 +71,187 @@ class Observable:
...
 
+    def commutes_with(self, obs: Observable):
+        """3M-TODO"""
+        # Naive version, just computing AB - BA, and compare to 0 matrix.
+        # TODO : distinguer si on a l'observable ou le pauli string
+        # TODO: traitement spécifique si observable diagonal ?
+
+        if self.is_diagonal:
+            if obs.is_diagonal:
+                return True
```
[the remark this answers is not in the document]
hJaffaliColibritd: Yes `commutes_with` return True if two observables commute. But the implementation is not finished; But we could add the return type of this method.

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -269,7 +426,7 @@ def __repr__(self) -> str:
         )
         shots = "" if self.shots == 0 else f", shots={self.shots}"
         label = "" if self.label is None else f", label={self.label}"
-        return f"ExpectationMeasure({self.observable}{targets}{shots}{label})"
+        return f"ExpectationMeasure({self.observables}{targets}{shots}{label})"
```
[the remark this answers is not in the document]
MoHermes: For this modified line, it should be `if len(self.observables) == 1`, otherwise, is being referenced before it has been assigned a value.

mpqp/core/instruction/measurement/expectation_value.py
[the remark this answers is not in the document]
JulienCalistoTD: Okay, I see. Just add a return value `boolean` in the function definition then.

mpqp/core/instruction/measurement/expectation_value.py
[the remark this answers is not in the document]
hJaffaliColibritd: Done in 68f2d29

mpqp/core/instruction/measurement/expectation_value.py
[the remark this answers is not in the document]
hJaffaliColibritd: Done by @MoHermes here c2b1c92, and completed here d72028c

mpqp/core/instruction/measurement/expectation_value.py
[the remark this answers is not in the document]
hJaffaliColibritd: Done

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -329,6 +330,20 @@ def to_matrix(self) -> Matrix:
             start=np.zeros((size, size), dtype=np.complex64),
         )
 
+    def commutes_with(self, other: PauliString) -> bool:
+        """
+        3M-TODO: Determine EFFICIENTLY if this pauli string is commuting with the one in parameter.
+        Args:
```
Henri-ColibrITD, line 336: the doc here is not done because of the TODO ? meaning it doesn't appear in the online doc, is that right ?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -809,8 +842,12 @@ class PauliStringMonomial(PauliString):
     def __init__(self, coef: Coef = 1, atoms: Optional[list["PauliStringAtom"]] = None):
         self.coef = coef
         """Coefficient of the monomial."""
-        self.atoms = [] if atoms is None else atoms
+        self._atoms = [] if atoms is None else atoms
+
+    @property
+    def atoms(self) -> list["PauliStringAtom"]:
```
Henri-ColibrITD, line 848: why moving the atoms to be a property ?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -912,6 +954,43 @@ def __hash__(self):
         atoms_as_tuples = tuple((atom.label for atom in self.atoms))
         return hash(atoms_as_tuples)
 
+    def commutes_with(self, other: PauliString) -> bool:
+        """Checks wether this Pauli monomial commutes (full commutativity) with the Pauli monomial in parameter. This
```
Henri-ColibrITD, line 958: why the "(full commutativity)" specification is there another meaning ?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -820,6 +857,11 @@ def nb_qubits(self) -> int:
     def monomials(self) -> list["PauliStringMonomial"]:
         return [PauliStringMonomial(self.coef, self.atoms)]
 
+    @property
+    def eigen_values(self) -> npt.NDArray[np.complex64]:
```
Henri-ColibrITD, line 861: isn't this always just a list of + and - 1s ?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -912,6 +954,43 @@ def __hash__(self):
...
+            >>> from mpqp.measures import I, X, Y, Z
+            >>> (I @ X @ Y).commutes_with(Z @ Y @ X)
+            True
+            >>> (X @ Z @ Z).commutes_with(Y @ Z @ I)
+            False
+            >>> (X @ Z @ Z).commutes_with(X @ Z @ Z)
+            True
+
+        """
+        if not isinstance(other, PauliStringMonomial):
```
Henri-ColibrITD, line 977: why do you take as input a PS if you only allow a monomial as input ?

Henri-ColibrITD: imo, you could also determine quite easily if a monomial commutes with a whole PS (you just have to check if it commutes with all the monomials individually)

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1004,6 +1083,9 @@ class PauliStringAtom(PauliStringMonomial):
     Args:
         label: The label representing the Pauli operator.
         matrix: The matrix representation of the Pauli operator.
+        eig_values: The eigenvalues associated with the Pauli operator.
+        eig_vectors: A list of eigenvectors associated with the Pauli operator.
+        basis_change: Basis change of the measurement associated with the Pauli operator.
```
Henri-ColibrITD, line 1088: it's weird to put this in the atom, I would put this in the measurement object instead no ?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1032,7 +1124,7 @@ def nb_qubits(self) -> int:
         return 1
 
     @property
-    def atoms(self):
+    def atoms(self) -> list["PauliStringAtom"]:
```
Henri-ColibrITD, line 1127: do you need to put explicit return type ?

mpqp/core/instruction/measurement/pauli_string.py
MoHermes: Yes, it's right.

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Yes, in a specific order depending on the atoms. You think we should store them as `int` or `bool` ?

mpqp/core/instruction/measurement/pauli_string.py
MoHermes: To keep the design consistent as it is also property in `PauliStringAtom`.

mpqp/core/instruction/measurement/pauli_string.py
MoHermes: Yes, it is always just a list of ±1s, because it is computed from the eigenvalues of Pauli matrices, which are always ±1.
And,  if we only care about the sign (not the actual values), we could store them as booleans.

mpqp/core/instruction/measurement/pauli_string.py
MoHermes: This to have `Pyright` to interpret the method consistently with it's parent class `PauliStringMonomial`

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -912,6 +954,43 @@ def __hash__(self):
...
+        """
+        if not isinstance(other, PauliStringMonomial):
+            raise ValueError(
+                f"Expected a PauliStringMonomial in parameter but got {type(other).__name__}"
+            )
+
+        if self.nb_qubits != other.nb_qubits:
+            raise NumberQubitsError(
+                f"The number of qubits of this Pauli monomial ({self.nb_qubits}) is not matching "
+                f"the one of the monomial in parameter ({other.nb_qubits})."
```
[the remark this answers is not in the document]
MoHermes: Good. It's more conciseand clearer.

examples/notebooks/3_Expectation_value_of_observables.ipynb
```diff
@@ -554,11 +554,176 @@
...
+   "cell_type": "code",
+   "execution_count": 22,
+   "id": "6d2571a4",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "BatchResult: 3 results\n",
```
hJaffaliColibritd, line 648: Update the notebook after we finished to replace BatchResult by Result

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: The implementation of the function itself is not finished indeed. That's why it was left like this. The goal is to optimize a bit more the method depending on which attribute is set (matrix, pauli_string, diag elements) to avoid conversions

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Pyright was complaining that atoms was an attribute in PauliStringMonomial and a property in PauliStringAtoms

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: You can for instance have qubit-wise commutativity, or full commutativity. The first one implies the second, but fully commuting observables are not necessary qubit-wise commuting. 

For example, XIY and XYI are qubit-wise commuting (you check if each pair of single pauli operator at each qubit index commute) and are fully commuting, but this is not the case of XXX and YYY, which are only fully commuting.

The different types of commutation lead to different grouping and measurement protocols to enhance the performance of expectation value evaluation

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Because it is a method inherited from PauliString, for which you can't override the type of the argument. For the moment we only needed to implement this for monomials, that is why we restricted to that case. We can maybe raise a NonImplementedError instead.

I think that you are right, but I wonder if there is any particular case

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: It was meant to be used to determine quickly the change of basis associated with a PauliStringMonomial, you just apply the change of basis for each qubit by iterating over the atoms

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1108,6 +1201,38 @@ def __eq__(self, other: object) -> bool:
     def __hash__(self):
         return hash(self.label)
 
+    def to_matrix(self) -> npt.NDArray[np.complex64]:
```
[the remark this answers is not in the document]
hJaffaliColibritd: I thought it was a convention to group all the duner methods together? Like grouping the properties together also?

hJaffaliColibritd (changes requested):

mpqp/core/instruction/measurement/pauli_string.py
Henri-ColibrITD: yup, `NonImplementedError` would be better

mpqp/core/instruction/measurement/pauli_string.py
MoHermes: Done here 6a5a0af

Henri-ColibrITD: Ok, I now have a tiny nitpick: now, the array doesn't represent eigen values anymore,  but rather something like negative eigen values. Also, in the commit you shared above, only one file is modified, does it mean that you didn't use this property anywhere ? (I might have the answer reading further, let's see)

mpqp/core/instruction/measurement/pauli_string.py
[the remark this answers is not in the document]
Henri-ColibrITD: once you read my reply, you can mark this as resolved

examples/notebooks/3_Expectation_value_of_observables.ipynb
Henri-ColibrITD: done ?

mpqp/core/instruction/measurement/pauli_string.py
Henri-ColibrITD: > As whether it's +1 or -1, I thought of True as representing +1 and False as representing -1

Ok, then the name would be something like `positive_eigen_values`

hJaffaliColibritd: > Ok, I now have a tiny nitpick: now, the array doesn't represent eigen values anymore, but rather something like negative eigen values. Also, in the commit you shared above, only one file is modified, does it mean that you didn't use this property anywhere ? (I might have the answer reading further, let's see)

Indeed we don't use it anywhere for the moment because it will be used for the manual multi-observable run for AWS/QLM, and was not used for IBM. We worked on both approaches in parallel in the same branch, but only a part of the second approach was actually implemented.  The eigen values of the PauliStringAtoms are supposed to help compute quickly the eigenvalues associated with a PauliMonomial by just taking the Kronecker product of the eigenvalues of the atoms.

examples/notebooks/3_Expectation_value_of_observables.ipynb
hJaffaliColibritd: Done in b051772.

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Done in 2a1aa65.

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Because with your current implementation @MoHermes, I think the result is false:
```Python
>>> from mpqp.measures import *
>>> (X @ Z @ Z).eigen_values
array([ True, False, False, False, False, False, False, False])
```

MoHermes: Check here 9158ad3

MoHermes: Ok will update the function name

hJaffaliColibritd: Done in dba3b3b

hJaffaliColibritd (approved):

PR #132, opened 2025-03-28
feat: qcircuit initializer

ThomasB-Colibri:
Added new feature : QCircuit.initializer using Qiskit's StatePreparation

Also added the from_other_language, only working with Qiskit and Qasm2 for now

Also added a random_statevector function in tools usefull to test the initializer

### DIFF
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -1389,19 +1389,11 @@ class QCircuit:
 
         elif isinstance(qcircuit, str):  # pyright: ignore[reportUnnecessaryIsInstance]
             lines = qcircuit.split('\n')
-            qasm_str = []
-            line_to_add = True
             for line in lines:
-                if "//" in line or line == '':
-                    line_to_add = False
-                if line_to_add is True:
-                    qasm_str.append(line)
-                else:
-                    line_to_add = True
-            qasm_str = '\n'.join(qasm_str)
-
-            if not qasm_str.startswith("OPENQASM 2.0"):
-                raise NotImplementedError(f"Error: only OpenQASM2 is supported for qasm external description of the circuit")
+                if not line.startswith("//") and line != '':
+                    if not line.startswith("OPENQASM 2.0"):
+                        raise NotImplementedError(f"Error: only OpenQASM2 is supported for qasm external description of the circuit")
+                    break
             return qasm2_parse(qcircuit)
         else:
             raise NotImplementedError(f"Error: {type(qcircuit)} is not supported.")


Henri-ColibrITD (changes requested):
a tiny comment, but appart from that it's good ! gg :D

tests/core/test_circuit.py
```diff
@@ -311,6 +333,23 @@ def test_to_other_language(
...
+            (random_circuit(None, 10), Language.QASM2),
+        ],
+)
+def test_from_other_language(circuit: QCircuit, language: Language):
+    circ_to_test = circuit.to_other_language(language)
+    qcircuit = QCircuit()
+    if isinstance(circ_to_test, QiskitCircuit):
+        qcircuit = QCircuit.from_other_language(circ_to_test)
+    elif isinstance(circ_to_test, str):
+        qcircuit = QCircuit.from_other_language(circ_to_test)
```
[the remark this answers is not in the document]
ThomasB-Colibri: i did it to avoid a flag for type errors for circ_to_test, since there are several entry types

mpqp/core/circuit.py
```diff
@@ -1322,6 +1332,69 @@ def to_other_language(
...
+            q_1: ─────┤ X ├
+                      └───┘
+        """
+        from mpqp.qasm.qasm_to_mpqp import qasm2_parse
+        from qiskit import QuantumCircuit
+
+        if isinstance(qcircuit, QuantumCircuit):
+            from qiskit import qasm2
+            
+            qasm2_code = qasm2.dumps(qcircuit)
```
hJaffaliColibritd, line 1386: In my experience, you have to remove the `reset` keyword from the OpenQASM2 code because it was not handled by our parser. Is this still true ? Or is it handled by the function `qasm2_parse` ?

mpqp/core/circuit.py
```diff
@@ -1322,6 +1332,69 @@ def to_other_language(
...
+            return qasm2_parse(qasm2_code)
+
+        elif isinstance(qcircuit, str):  # pyright: ignore[reportUnnecessaryIsInstance]
+            if not "OPENQASM 2.0;" in qcircuit:
+                raise NotImplementedError(f"Error: only open QASM2 is supported")
```
hJaffaliColibritd, line 1391:
```suggestion
                raise NotImplementedError(f"Error: only OpenQASM2 is supported for qasm external description of the circuit.")
```

mpqp/tools/circuit.py
```diff
@@ -92,6 +93,32 @@ def random_circuit(
     return qcircuit
 
 
+def random_statevector(
```
hJaffaliColibritd, line 96: I would name it `random_statevector_from_random_circuit` (a bit too long, I confess), or maybe precise the method to generate this random state in parameter, because it may not be obvious that it is done using the random circuit feature. 

(one can randomly sample a vector in the hilbert space, one can randomly generate a n-qubit unitary, or other ways of generating random states)

mpqp/tools/circuit.py
```diff
@@ -92,6 +93,32 @@ def random_circuit(
...
+        nb_qubits: int = 5, 
+        seed: Optional[int] = None,
+) -> npt.NDArray[np.complex64]: 
+    """
+    This function creates a statevector with a specified number of qubits.
```
hJaffaliColibritd, line 101: Preise in the doc that the statevector is generated via a random circuit and after execution on IBM AER simulator. Interesting to know that we did an arbitrary choice of local device.

tests/core/test_circuit.py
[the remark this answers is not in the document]
hJaffaliColibritd: Then you better use a synthax of the form 
```
if TYPE_CHECKING:
    assert isinstance(circ_to_test, (QiskitCircuit, str))
qcircuit = QCircuit.from_other_language(circ_to_test)
```

If it is only to satisfy to typechecker, then just add this `if` like this.
Does it make sense ?

tests/core/test_circuit.py
```diff
@@ -311,6 +333,23 @@ def test_to_other_language(
...
+@pytest.mark.parametrize(
+        "circuit, language",
+        [
+            (random_circuit(None, 10), Language.QISKIT),
+            (random_circuit(None, 10), Language.QASM2),
+        ],
+)
+def test_from_other_language(circuit: QCircuit, language: Language):
+    circ_to_test = circuit.to_other_language(language)
+    qcircuit = QCircuit()
```
hJaffaliColibritd, line 345: This line is useless right ?

hJaffaliColibritd (changes requested):

mpqp/core/circuit.py
ThomasB-Colibri: I used directly the StatePreparation function and not the initialize, since it is composed of a reset and a StatePreparation. So we don't have the reset token here.
But yes the reset token isn't handled by qasm2_parse. I did something earlier to skip the token but since it wasn't useful I deleted it.

mpqp/tools/circuit.py
ThomasB-Colibri: I propose `statevector_from_random_circuit`

tests/core/test_circuit.py
ThomasB-Colibri: Yes, this line isn't used anymore since the modifications of the last comment.

mpqp/core/circuit.py
hJaffaliColibritd: Ok good. i resolve the conversation

tests/core/test_circuit.py
```diff
@@ -336,17 +352,18 @@ def test_to_other_language(
...
         ],
 )
 def test_from_other_language(circuit: QCircuit, language: Language):
     circ_to_test = circuit.to_other_language(language)
-    qcircuit = QCircuit()
-    if isinstance(circ_to_test, QiskitCircuit):
-        qcircuit = QCircuit.from_other_language(circ_to_test)
-    elif isinstance(circ_to_test, str):
-        qcircuit = QCircuit.from_other_language(circ_to_test)
+    assert(isinstance(circ_to_test, (QiskitCircuit, str)))
```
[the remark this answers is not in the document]
hJaffaliColibritd: So I would really just use the piece of code proposed in another comment (I resolved the other comment, so we focus only on this one); 
```
if TYPE_CHECKING:
    assert isinstance(circ_to_test, (QiskitCircuit, str))
qcircuit = QCircuit.from_other_language(circ_to_test)
```

mpqp/tools/circuit.py
hJaffaliColibritd: Ok for me.

tests/core/test_circuit.py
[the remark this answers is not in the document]
ThomasB-Colibri: oh okay, I didn't understood how to use it at first, is it ok this way ?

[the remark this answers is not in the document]
hJaffaliColibritd: Very good. I resolve the conversation

hJaffaliColibritd (approved):

JulienCalistoTD (approved):

tests/core/test_circuit.py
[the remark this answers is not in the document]
hJaffaliColibritd: I resolve then, solved in another comment

mpqp/core/circuit.py
```diff
@@ -1322,6 +1332,69 @@ def to_other_language(
...
+
+        if isinstance(qcircuit, QuantumCircuit):
+            from qiskit import qasm2
+
+            qasm2_code = qasm2.dumps(qcircuit)
+            return qasm2_parse(qasm2_code)
+        
+        elif isinstance(qcircuit, str):  # pyright: ignore[reportUnnecessaryIsInstance]
+            if not "OPENQASM 2.0;" in qcircuit:
+                raise NotImplementedError(f"Error: only OpenQASM2 is supported for qasm external description of the circuit")
```
MoHermes, line 1392: Here to use `if not qcircuit.startswith("OPENQASM 2.0"):` instaed of hardcoding it IF it's always in the begining of the string, as it is faster!
and also I have a suggestion @hJaffaliColibritd @ThomasB-Colibri, If we plan to support more OpenQASM versions in the future, a regex check is recommended?

ThomasB-Colibri: for the regex check, I think we can do this later since we will need to extend the code for other versions anyway

mpqp/tools/circuit.py
```diff
@@ -92,6 +93,34 @@ def random_circuit(
...
+    
+    Examples:
+        >>> print(statevector_from_random_circuit(2, seed=123)) # doctest: +NORMALIZE_WHITESPACE
+        [0.70710678+0.j 0.        +0.j 0.70710678+0.j 0.        +0.j]
+    """
+    from mpqp.execution import run, IBMDevice, Result
+
+    mpqp_circ = random_circuit(None, nb_qubits, None, seed)
+    res = run(mpqp_circ, IBMDevice.AER_SIMULATOR_STATEVECTOR)
+    assert isinstance(res, Result)
```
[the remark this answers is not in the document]
hJaffaliColibritd: If done, resolve the comment @MoHermes

mpqp/core/circuit.py
hJaffaliColibritd: > I made a proposal in mp (help, I don't know how to right clear code here)
> 
> Since the first line can be a comment, it parses it until the first qasm code line and checks if it's is the good one. But I don't know if it is really faster

To write clean code use put the code between triple backquotes " ``` "

Then maybe we remove all the comments from the file before parsing it ? Because if it is not a comment, and OPENQASM2.0 is the last line of the file, I agree with Momo that it can cause a problem

ThomasB-Colibri: @hJaffaliColibritd @MoHermes I made a modification that checks if the are comments on the string and removes them, then if OPENQASM 2.0 is the first instruction

mpqp/core/circuit.py
```diff
@@ -1389,11 +1389,19 @@ def from_other_language(
...
         elif isinstance(qcircuit, str):  # pyright: ignore[reportUnnecessaryIsInstance]
             lines = qcircuit.split('\n')
+            qasm_str = []
+            line_to_add = True
             for line in lines:
-                if not "//" in line and line != '':
-                    if not "OPENQASM 2.0" in line:
-                        raise NotImplementedError(f"Error: only OpenQASM2 is supported for qasm external description of the circuit")
-                    break
+                if "//" in line or line == '':
```
hJaffaliColibritd, line 1395: I would use maybe the `replace` method of a string to replace any string starting with '\\' until jump of line by the empty string '' ?

hJaffaliColibritd (changes requested):

mpqp/core/circuit.py
hJaffaliColibritd: With the last commit I agree with the logic, that we go through the document, we discard the checking of comment lines, and check for the first code line that it starts with OPENQASM 2.0

ThomasB-Colibri: It is fixed, I check line by line until the first openqasm instruction and check if it starts with OPENQASM 2

hJaffaliColibritd (approved):

mpqp/core/circuit.py
hJaffaliColibritd: Done in 422af9c. I resolve

MoHermes (approved):

PR #135, opened 2025-04-01
feat: add submit_job_azure function for job submission on Azure devices

JulienCalistoTD:


mpqp/execution/providers/azure.py
```diff
@@ -70,6 +71,53 @@ def run_azure(job: Job) -> Result:
     return result
 
 
+@typechecked
+def submit_job_azure(job: Job) -> tuple[str, "AzureQuantumJob"]:
```
hJaffaliColibritd, line 75: You should call this function in run (to not duplicate the same code), and then wait for the result in run (sync execution), like we already do with `job_sim.result()`. But here the `job_sim` should come from your `submit_job_azure`

hJaffaliColibritd (changes requested):

mpqp/execution/providers/azure.py
JulienCalistoTD: done: [1bd7383](https://github.com/ColibrITD-SAS/mpqp/pull/135/commits/1bd73835c83df7c9e3c7ad807db435f8761bf334)

MoHermes: I was having the same comment as Hamza, but it's already done.

hJaffaliColibritd (approved):

MoHermes (approved):

.github/workflows/tests.yml
```diff
@@ -73,6 +77,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
+      - name: Install osqp
+        run: |
+            pip install --upgrade pip
+            pip install "osqp==1.0.1"
```
hJaffaliColibritd, line 83: we can't force this version in the requirements ? after having installed `azure-quantum` ?

hJaffaliColibritd (approved):

.github/workflows/tests.yml
JulienCalistoTD: is not from `azure-quantum` but i move it to the requirements.

PR #138, opened 2025-04-04
Fix symbolic print customgate

Henri-ColibrITD:


mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -82,12 +85,33 @@ def to_other_language(
         self,
         language: Language = Language.QISKIT,
         qiskit_parameters: Optional[set["Parameter"]] = None,
+        printing: bool = False,
     ):
+        print(printing)
```
hJaffaliColibritd, line 90: This should be removed (that is why the tests don't pass)

hJaffaliColibritd:
```suggestion
```

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -82,12 +85,33 @@ def to_other_language(
...
+                    assert isinstance(symbol, Expr)
+                _qiskit_parameter_adder(symbol, qiskit_parameters)
+
+            if printing:
+                from qiskit import QuantumCircuit
+
+                dummy_circuit = QuantumCircuit(self.nb_qubits)
+                for param in qiskit_parameters:
+                    dummy_circuit.rx(param, 0)
+                return dummy_circuit.to_gate(label="CustomGate")
```
hJaffaliColibritd, line 114: Is this meant to be removed after finishing the manual tests ?

hJaffaliColibritd (commented):
Should we add some tests related to the problem tat happened ? 

or some examples for doctest ?

mpqp/core/instruction/gates/custom_gate.py
Henri-ColibrITD: no, basically the problem came from the fact that you cannot create qiskit instructions from matrix with variables in the matrix, but we need that for printing, so the solution I chose is that for printing we use a dummy gate with all the symbolic parameters as input

hJaffaliColibritd: Ok I understand now why we have some `rx` gates.

mpqp/core/circuit.py
```diff
@@ -1076,6 +1077,9 @@ def to_other_language(
                 translation issues. if True, a warning will be raised.
             skip_pre_measure: If true, the ``pre_measure`` circuit will not be
                 added to the output.
+            printing: If printing dummy gates will replace custom gates (because
```
hJaffaliColibritd, line 1080: If `True` .... to match the discussion in the other PR ?

hJaffaliColibritd (approved):

MoHermes (approved):
Just added one comment, but clear work.

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -82,12 +85,38 @@ def to_other_language(
...
+                    raise ValueError(
+                        "Custom gates defined with symbolic variables cannot be"
+                        " exported to qiskit."
+                    )
+                from qiskit import QuantumCircuit
+
+                dummy_circuit = QuantumCircuit(self.nb_qubits)
+                for param in qiskit_parameters:
+                    dummy_circuit.rx(param, 0)
+                return dummy_circuit.to_gate(label="CustomGate")
```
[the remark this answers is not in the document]
Henri-ColibrITD: The fact that the circuit is called `dummy_circuit` isn't enough ?
(Here is the definition of dummy I'm referring to:
> A deliberately nonfunctional device or tool used in place of a functional one.

)

PR #139, opened 2025-04-07
Feat multi observable cirq

MoHermes:


test_multiobs.ipynb
hJaffaliColibritd, line 1: Remove this whole file from the PR, it should be in your .gitignore

mpqp/execution/providers/google.py
```diff
@@ -253,21 +263,34 @@ def run_local_processor(job: Job) -> Result:
...
+                language=Language.CIRQ, circuit=cirq_circuit
+            )
+
+            if isinstance(translated, Cirq_PauliSum):
+                if TYPE_CHECKING:
+                    assert isinstance(translated, list)
+                    assert all(
+                        isinstance(item, Cirq_PauliString) for item in translated
+                    )
+                translated = next(iter(translated), None)
```
hJaffaliColibritd, line 278: What this is useful for ?

mpqp/execution/providers/google.py
```diff
@@ -125,7 +124,8 @@ def run_local(job: Job) -> Result:
         )
 
     from cirq.circuits.circuit import Circuit as CirqCircuit
-    from cirq.ops.pauli_string import PauliString as CirqPauliString
+
+    # from cirq.ops.pauli_string import PauliString as CirqPauliString
```
hJaffaliColibritd, line 128: We don't need to import this anymore? If so, then remove the comment completely ?

hJaffaliColibritd (changes requested):

mpqp/execution/providers/google.py
MoHermes: Yes, I just pushed the latest version, then I do the clean-up

mpqp/execution/providers/google.py
MoHermes: `translated` is `Cirq_PauliSum` at runtime. `next(iter(` gets the first item from the collection, which is iterable.
It could be cirq-specific type, that's why when used list it was misleading and got errors, but still looking to make it works

PR #140, opened 2025-04-10
Fix patch before release

hJaffaliColibritd:


mpqp/core/circuit.py
```diff
@@ -1485,13 +1485,17 @@ def to_other_device(
...
 
+                if TYPE_CHECKING:
+                    assert isinstance(device, IBMDevice)
+
                 backend = get_backend(device)
                 pm = generate_preset_pass_manager(backend=backend, optimization_level=1)
                 qiskit_circuit = pm.run(qiskit_circuit)
 
+            # TODO his was added to effectively stored the transpiled circuit, double check
+            self.transpiled_circuit = qiskit_circuit
```
JulienCalistoTD, line 1498: don't need this for now

mpqp/core/circuit.py
```diff
@@ -1543,6 +1547,9 @@ def to_other_device(
                 )
 
                 cirq_device.validate_circuit(cirq_circuit)
+
+            # TODO his was added to effectively stored the transpiled circuit, double check
+            self.transpiled_circuit = cirq_circuit
```
JulienCalistoTD, line 1552: same here

mpqp/core/circuit.py
hJaffaliColibritd: Ok then remove both and resolve

mpqp/core/circuit.py
hJaffaliColibritd: Remove then and resolve

PR #142, opened 2025-04-15
Feat unitary decomposition

MathieuG-Colibri:
Feature : Decomposition of any unitary matrices using quantum Shannon decomposition.

Fix : The global phase is now taken into account with the to_matrix() method (test added) and controlled gate cannot have the same control/target qubit(s) in case of a negative index (test also added).


mpqp/core/circuit.py
```diff
@@ -287,6 +287,19 @@ def _check_components_targets(self, components: Instruction | NoiseModel):
                 raise ValueError(
                     "In noisy circuits, BasisMeasure must span all qubits in the circuit."
                 )
+        if isinstance(components, ControlledGate):
+            for i in range(len(components.controls)):
+                components.controls[i] += (
+                    self.nb_qubits if components.controls[i] < 0 else 0
```
JulienCalistoTD: Check whether negative `controls` and `targets` are handled for every gate, and consider finding a way to support them for all gates not just `ControlledGate`

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -8,13 +8,15 @@
 
 from typeguard import typechecked
 
+# from mpqp.core.circuit import QCircuit
```
JulienCalistoTD: remove this line
```suggestion
```

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,185 @@
...
+    Returns the gray code of n.
+    """
+    return n ^ (n >> 1)
+
+
+def unitary_SVD(U: Matrix) -> tuple[Matrix, Matrix, Matrix]:
+    """
+    Returns the decomposition of the unitary using eigenvalues decomposition.
+    The goal of this SVD is to have following result : U = (I ⊗ V) @ D @ (I ⊗ W).
+    D is in the form of a multiplexed Rz operator.
```
JulienCalistoTD, line 28: Specify the algorithm/paper in the doc

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,185 @@
...
+
+    This function should be used only in the context of the Quantum Shannon Decomposition.
+
+    Args:
+        U : The unitary matrix to decompose.
+
+    Returns:
+        3 matrices
+    """
+    length = len(U)
```
JulienCalistoTD, line 47: If you're using variable names from a paper, add comments to make them easier to understand. Otherwise, use more descriptive names

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,185 @@
...
+    The goal of this SVD is to have following result : U = (I ⊗ V) @ D @ (I ⊗ W).
+    D is in the form of a multiplexed Rz operator.
+
+    This function should be used only in the context of the Quantum Shannon Decomposition.
+
+    Args:
+        U : The unitary matrix to decompose.
+
+    Returns:
+        3 matrices
```
JulienCalistoTD: What are the 3 matrices supposed to be?

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,185 @@
...
+def _decompose(U: Matrix, circuit: QCircuit, position: int = 0) -> QCircuit:
+    """
+    This function recursively decompose the matrix U into the circuit then returns it.
+
+    For 1 qubit operators it executes a ZYZ decomposition.
+    For higher dimensions it does a Quantum Shannon decomposition.
+    """
+    if len(U) == 2:  # Decompose a 1 qubit operator
+        delta = np.angle(np.linalg.det(np.asarray(U, dtype=np.complex128))) / len(U)
+        V = U / np.exp(1j * delta)  # extract the global phase so that V is SU
```
JulienCalistoTD: Same for variable naming

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def _decompose(U: Matrix, circuit: QCircuit, position: int = 0) -> QCircuit:
+    """
+    This function recursively decompose the matrix U into the circuit then returns it.
+
+    For 1 qubit operators it executes a ZYZ decomposition.
+    For higher dimensions it does a Quantum Shannon decomposition.
+    """
+    if len(U) == 2:  # Decompose a 1 qubit operator
```
JulienCalistoTD, line 135: Did you check that `U` has the correct size?

tests/core/instruction/gates/test_custom_gate.py
```diff
@@ -145,3 +146,27 @@ def test_custom_gate_with_random_circuit(circ_size: int, device: AvailableDevice
     assert isinstance(result1, Result)
     assert isinstance(result2, Result)
     assert matrix_eq(result1.amplitudes, result2.amplitudes, 1e-4, 1e-4)
+
+
+def test_decomposition_short():
+    from scipy.stats import unitary_group
+
+    for i in range(1, 4):
+        u = np.asarray(unitary_group.rvs(2**i), dtype=np.complex128)
```
JulienCalistoTD: You have a function in the tool to create a `random_unitary_matrix`

tests/core/test_circuit.py
```diff
@@ -453,6 +454,12 @@ def test_instruction_no_target(component: Instruction | NoiseModel):
         assert qubits == noise.targets
 
 
+def test_invalid_controlled_gate():
+    circuit = QCircuit(3)
+    with pytest.raises(ValueError):
+        circuit.add(CNOT(0, -3))
```
JulienCalistoTD: test with all gates and also for target

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: Normally U is called from a customGate so it's size should always have a 2**n size.

But maybe this is a change I can make.

mpqp/core/circuit.py
MathieuG-Colibri: The issue is raised only for the Toffoli gate so my interpretation was that the problem wasn't checked if we had a 3 qubit gate or more.
So that's why I've put it there.

tests/core/test_circuit.py
MathieuG-Colibri: I have changed the test with the Toffolli which was the only gate where this happened.

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: Done [9ca5408](https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/9ca5408f7abfe72a37bd0cadaff69eeab8683772)

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: Done [9ca5408](https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/9ca5408f7abfe72a37bd0cadaff69eeab8683772)

mpqp/core/instruction/gates/custom_gate.py
MathieuG-Colibri: Done [9ca5408](https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/9ca5408f7abfe72a37bd0cadaff69eeab8683772)

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: Done [9ca5408](https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/9ca5408f7abfe72a37bd0cadaff69eeab8683772)

mpqp/tools/generics.py
```diff
@@ -60,7 +60,10 @@
 
 """
 Matrix = Union[
-    npt.NDArray[np.complex64], npt.NDArray[np.float64], npt.NDArray[np.object_]
+    npt.NDArray[np.complex64],
+    npt.NDArray[np.float64],
+    npt.NDArray[np.object_],
+    npt.NDArray[np.complex128],
```
MoHermes, line 66: Is it better to generalize here `Matrix = npt.NDArray[np.typing.NDArray[Any]]` ?? not so sure!!

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+        A tuple of 3 matrices (V,D,W) so that U = (I ⊗ V) @ D @ (I ⊗ W)
+
+    Reference:
+    .. [1] V. Shende, S. S. Bullock, and I. Markov, “Synthesis of quantum logic
+        circuits,” Computer-Aided Design of Integrated Circuits and Systems,
+        IEEE Transactions on, vol. 25, pp. 1000 – 1010, July 2006.
+    """
+
+    # Decompose U into it's 2 submatrices G0 and G1
+    length = len(U)
```
MoHermes, line 47: Add size check for `U` ?

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+    For higher dimensions it does a Quantum Shannon decomposition.
+
+    Reference:
+    .. [1] Mikko Möttönen, Juha J. Vartiainen, Ville Bergholm, and Martti M. Salomaa. 2004. Quantum circuits for general multi-qubit gates. American Physical Society (APS) : 93-13.
+    """
+    if len(U) == 2:  # Decompose a 1 qubit operator
+        delta = np.angle(np.linalg.det(np.asarray(U, dtype=np.complex128))) / len(U)
+
+        # extract the global phase so that SU is a special unitary, note that if U is a special unitary then delta = 0
+        SU = U / np.exp(1j * delta)
```
MoHermes, line 139: Rename `SU` to something descriptive!

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+    else:  # 2 qubits or more
+        length = len(U)
+        U12, MuxRy, V12 = cossin(U, p=length // 2, q=length // 2, separate=False)
+
+        # Extracts the rotations of the multiplexed Ry for later decomposition
+        thetas = []
+        for i in range(MuxRy.shape[0] // 2):
+            thetas.append(np.arccos(MuxRy[i][i]))
+        thetas = np.array(thetas)
+        # checks typing because of cossin U12 and V12 could be tuples themselves
```
MoHermes: I feel we don't need to add comments for every step!!

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+        # Extracts the rotations of both multiplexed Rz for later decomposition
+        du = np.angle(MuxRzu.diagonal())
+        for i in range(len(du) // 2):
+            du[i] *= -1
+
+        dv = np.angle(MuxRzv.diagonal())
+        for i in range(len(dv) // 2):
+            dv[i] *= -1
+
+        # Now recursively decompose every obtained matrices.
```
MoHermes, line 173: Same here we don't need to add comments!!

tests/core/instruction/gates/test_custom_gate.py
MathieuG-Colibri: I could either make a function that:
- computes the expodential of a random hermitian
- simply call the scipy function
This solution seems the best imo because it removes typing issues and is faster than the other method.

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -163,9 +164,11 @@ def __repr__(self) -> str:
         label = f", \"{self.label}\"" if self.label else ""
         return f"CustomGate({UnitaryMatrix(self.matrix)}, {self.targets}{label})"
 
-    def decompose(self):
+    def decompose(self) -> "QCircuit":
```
hJaffaliColibritd, line 167: I would add some examples in the Docstring, to show what it gives as a circuit.

And I would also add this wherever we use `CustomGate` in the examples scripts or notebooks.

mpqp/tools/generics.py
hJaffaliColibritd: Why not `np.ndarray` in that case?

mpqp/tools/unitary_decomposition.py
Henri-ColibrITD: I would write the expression in math format (latex formula)

mpqp/tools/generics.py
Henri-ColibrITD: I think your type is messed up @MoHermes it should read `Matrix = npt.NDArray[Any]` no ?

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
+from mpqp.core.circuit import QCircuit
+from mpqp.tools import Matrix
+import math
+from mpqp.gates import CNOT, Ry, Rz
+from scipy.linalg import cossin
+import numpy as np
+
+
+def _gray_code(n: int):
+    """
```
Henri-ColibrITD: no need for docstring for private functions, especially when they are that simple, I think

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+    Returns the decomposition of a multiplexed Rz or Ry gate.
+    The circuit is composed of a succession of CNOTs and rotations according to the original operator.
+
+    This function is originally from cirq's implementation.
+    Args:
+        thetas : A list of floats that are the rotations to be applied on each qubits.
+        circuit: The circuit in which the decomposition is stocked.
+        position : On which qubit is the rotation is taking place.
+        rotation : A string containing either Ry or Rz in the case of the Shannon decomposition.
```
Henri-ColibrITD: instead of str type, your rotation could be `Litteral[Ry,Rz]` to avoid allowing typos

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+
+        angle *= 2 / len(thetas)
+        # CNOT's control is the changed bit of two consecutive numbers in gray code
+        changed = _gray_code(i) ^ _gray_code(i + 1)
+        control = next(i for i in range(len(thetas)) if (changed >> i & 1))
+        control = max(-control - position - 1 + circuit.nb_qubits, 1)
+        if np.abs(angle) > 1e-9:  # Dodge unnecessary rotations
+            (
+                circuit.add(Ry(angle, position))
```
Henri-ColibrITD: if you follow my previous proposition, then you can replace this ternary by `circuit.add(rotation(angle, position))`, which seems much nicer to me

mpqp/tools/unitary_decomposition.py
Henri-ColibrITD: same minus the exclamation marks ^^

Although for implementations from papers one could argue this is more valuable than usual comments

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+
+        return circuit
+
+
+def quantum_shannon_decomposition(U: Matrix) -> QCircuit:
+    """
+    Returns a circuit containing the decomposition of a unitary.
+    The resulting circuit is composed of gates CNOT, Ry and Rz.
+
+    The Quantum Shannon Decomposition works by splitting an unitary into 3 matrices using the CSD (cosine sine decomposition).
```
Henri-ColibrITD: we try to limit the length of comments lines, even though this is not enforced

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+
+    This process repeats until the matrices are reduced to 1 qubit gates which can be easily split with the ZYZ decomposition.
+
+    Args:
+        U: The unitary matrix to be decomposed
+
+    Returns:
+        A quantum circuit containing the decomposition of U.
+
+    References:
```
Henri-ColibrITD, line 232: ? ^^

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Examples:
+        >>> from mpqp.tools.unitary_decomposition import quantum_shannon_decomposition
```
Henri-ColibrITD: those imports are not needed, ask @JulienCalistoTD for more details

tests/core/instruction/gates/test_controlled_gate.py
```diff
@@ -38,6 +38,11 @@ def test_gate_repr(gate: Gate, expected_repr: str) -> None:
     assert repr(gate) == expected_repr
 
 
+def test_negative_index_toffoli():
+    with pytest.raises(ValueError):
+        TOF([0, -1], 1)
```
Henri-ColibrITD: is the problem only with toffoli ? should this test be more generic ?

Henri-ColibrITD (changes requested):
a few small remarks but otherwise good job !!

tests/core/instruction/gates/test_controlled_gate.py
MathieuG-Colibri: The problem happens with a multi controlled gate (and a multi target controlled gate).
So in the case of our native gates yes it only happens with the Toffoli.

Also since I've wanted to implement the creation of custom controlled gate then the fix will work here also.

mpqp/core/circuit.py
```diff
@@ -782,8 +782,7 @@ def to_matrix(self) -> npt.NDArray[np.complex64]:
         matrix = Operator.from_circuit(qiskit_circuit).reverse_qargs().to_matrix()
         if TYPE_CHECKING:
             assert isinstance(matrix, np.ndarray)
-
-        return matrix
+        return matrix * np.exp(1j * self.gphase)
```
hJaffaliColibritd:
```suggestion
        if self.gphase != 0:
            matrix *= np.exp(1j * self.gphase)
        return matrix
```

mpqp/core/instruction/gates/controlled_gate.py
```diff
@@ -32,10 +32,17 @@ def __init__(
...
         if len(set(controls).intersection(set(targets))):
             raise ValueError(
                 f"Common registers between targets {targets} and controls {controls}"
             )
+        for control in controls:
+            if control < 0:
+                raise ValueError(f"Negative index in controls: {control}")
+        for target in targets:
+            if target < 0:
+                raise ValueError(f"Negative index in targets: {target}")
```
hJaffaliColibritd:
```suggestion
        if not all([control >= 0 for control in controls]):
            raise ValueError(f"Negative index in controls: {controls}")
        if not all([target>= 0 for target in targets]):
            raise ValueError(f"Negative index in targets: {targets}")
```

hJaffaliColibritd (changes requested):

mpqp/tools/generics.py
hJaffaliColibritd: I agree with @Henri-ColibrITD

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    Returns the gray code of n.
+    """
+    return n ^ (n >> 1)
+
+
+def _unitary_SVD(U: Matrix) -> tuple[Matrix, Matrix, Matrix]:
+    r"""
+    Returns the decomposition of the unitary using eigenvalues decomposition.
+    The goal of this SVD is to have following result : U = (I ⊗ V) @ D @ (I ⊗ W).
```
hJaffaliColibritd: Here the `@` is supposed to refer to a matrix product ? Usually `@` refers to tensor product. I would replace it with  a `\times` in LaTeX or just `.` or `x` instead

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+def _gray_code(n: int):
+    """
+    Returns the gray code of n.
+    """
+    return n ^ (n >> 1)
+
+
+def _unitary_SVD(U: Matrix) -> tuple[Matrix, Matrix, Matrix]:
+    r"""
+    Returns the decomposition of the unitary using eigenvalues decomposition.
```
hJaffaliColibritd, line 23: Does it work for any arbitrary `U`, or there are some conditions?

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+    The whole equation should be in the form :
+    `U = \begin{pmatrix}V & 0 \\ 0 & V end{pmatrix} \times \begin{pmatrix}D & 0 \\ 0 & D^{\dagger} \end{pmatrix} \times \begin{pmatrix}W & 0 \\ 0 & W \end{pmatrix}`
+
+    In the context of the quantum Shannon decomposition U is the result of a cosine sine decomposition, hence U is always in the form : `U =  \begin{pmatrix}U_0 & 0\\ 0 & U_1 \end{pmatrix}`
+
+    Args:
+        U : The unitary matrix to decompose.
+
+    Returns:
+        A tuple of 3 matrices (V,D,W) so that U = (I ⊗ V) @ D @ (I ⊗ W)
```
hJaffaliColibritd: Same as above

hJaffaliColibritd (changes requested):

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+    Reference:
+    .. [1] V. Shende, S. S. Bullock, and I. Markov, “Synthesis of quantum logic
+        circuits,” Computer-Aided Design of Integrated Circuits and Systems,
+        IEEE Transactions on, vol. 25, pp. 1000 – 1010, July 2006.
+    """
+
+    # Decompose U into it's 2 submatrices G0 and G1
+    length = len(U)
+
+    # U is composed of 2 matrices on it's diagonal here we need to extract them into G0 and G1
```
hJaffaliColibritd:
```suggestion
    # U is a block diagonal matrix. Here we the two blocks in G0 and G1
```

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+    g0 = []
+    g1 = []
+    for i in range(length // 2):
+        g0.append(list(SU[i][0 : length // 2]))
+        g1.append(list(SU2[i][length // 2 : length]))
+
+    G0 = np.array(g0)
+    G1 = np.array(g1)
+
+    # Build G as G = V @ D² @ V†
```
hJaffaliColibritd, line 61: Rewrite the comment with the code variables names, otherwise rename the variables, because here you don't use V neither D before line 56.

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: Or `Literral[type[Ry], type[Rz]]` ?
I would even put a boolean, if only two values are allowed and the argument mandatory

hJaffaliColibritd (changes requested):

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: SU is short for special unitary which is a common group in these kind of papers.
It's the same appelation as U but it's a less known group.

I don't really know what to rename it to in this case...

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: No need I already check the size at the beginning of the process. If the unitary is of size 2**n then the rest is guaranteed

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+            -angle if bin(_gray_code(i) & j).count('1') % 2 else angle
+            for j, angle in enumerate(thetas)
+        )
+
+        angle *= 2 / len(thetas)
+        # CNOT's control is the changed bit of two consecutive numbers in gray code
+        changed = _gray_code(i) ^ _gray_code(i + 1)
+        control = next(i for i in range(len(thetas)) if (changed >> i & 1))
+        control = max(-control - position - 1 + circuit.nb_qubits, 1)
+        if np.abs(angle) > 1e-9:  # Dodge unnecessary rotations
```
hJaffaliColibritd: Shouldn't we use a global variable for that so we can adjust the precision if one day we have a problem?

mpqp/tools/unitary_decomposition.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    This function recursively decompose the matrix U into the circuit then returns it.
+
+    For 1 qubit operators it executes a ZYZ decomposition.
+    For higher dimensions it does a Quantum Shannon decomposition.
+
+    Reference:
+    .. [1] Mikko Möttönen, Juha J. Vartiainen, Ville Bergholm, and Martti M. Salomaa. 2004. Quantum circuits for general multi-qubit gates. American Physical Society (APS) : 93-13.
+    """
+    if len(U) == 2:  # Decompose a 1 qubit operator
```
hJaffaliColibritd, line 135: In this case, since you laready have a single qubit unitary, why introducing 3 gates, global phase and so on, and not just using the `U` gate of mpqp directly ? Simpler, even in the execution workflow.

And this will directly do the trick for what we were discussing on teams, to avoid having 3 gates + additional gphase in the circuit, since you call this function recursively, right ?

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: I agree to let the comments when we follows steps from the article, it helps when we need to come back to it, or can be used by external users as it is

tests/core/instruction/gates/test_custom_gate.py
hJaffaliColibritd: Then maybe package your ideas in a single `random_unitary_matrix` function in `mpqp.tools.maths` and add an argument to select the mode of generation of the unitary (if it is a tensor product of local unitaries, a global unitary (exp of H or from scipy), ...)

hJaffaliColibritd (changes requested):

mpqp/tools/generics.py
MoHermes: Yeah yeah, I mistakenly did it twice but you got my idea, haha

mpqp/tools/unitary_decomposition.py
MoHermes: Ok as it is refrenced in the paper.

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: It's not possible to have Literal[type[Ry]] because it's not a value per say but I did change the typing to type[Rz] | type[Ry] to be able to directly call the constructor in the function.

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: G is used to define V and D in this case so I think it makes sense for the comment to be here since we define D and V two lines after.

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: Yes it would be easier but that wasn't the goal of the feature no ?
The goal was to decompose any unitary into the computationnal basis Rx, Ry, Rz, CNOT.

We can rewrite the function to avoid decomposition for 1 qubit gates in the case of big gates yes but otherwise isn't it just counterproductive to the goal of the function ?

But yes I agree the inherent complexity of the circuit scales so quickly it would be nice to simplify it a bit.

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/core/instruction/gates/controlled_gate.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/core/circuit.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done... I think ? https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/core/instruction/gates/custom_gate.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/8ba6a9469793250bb1045d6c917d111ad8a0548a

mpqp/core/circuit.py
hJaffaliColibritd: So what is the last status of this issue?
The code was moved to ToffoliGate class ? or to ControlledGate in general ?

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: Was the mathematical expression here updated in latex format ?

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: Ok great, for the optimization into a unique U gate, I'll add a trello task.

Resolved

mpqp/core/circuit.py
hJaffaliColibritd: @MathieuG-Colibri  ?

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: @MathieuG-Colibri ?

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: Is this solved ? @MathieuG-Colibri  ?

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: @MoHermes  if good for you can you resolve the conversation ?

mpqp/core/circuit.py
MathieuG-Colibri: Fixed in CustomControlledGate branch

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: Done

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: Done

mpqp/core/circuit.py
hJaffaliColibritd: @JulienCalistoTD  then I resolve

tests/core/instruction/gates/test_custom_gate.py
hJaffaliColibritd: This was completed @MathieuG-Colibri or still work in progress?

mpqp/tools/generics.py
hJaffaliColibritd: @MathieuG-Colibri  did you take into account this comment ?

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: @MathieuG-Colibri  ?

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: Indeed, you right

mpqp/tools/unitary_decomposition.py
MoHermes: I resolve.

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: I added precision to the documentation in https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/ed43b21b5d3c902170d960dc17cc3fd00c4e164e

tests/core/instruction/gates/test_controlled_gate.py
MathieuG-Colibri: Moved the test to the custom controlled gate branch

tests/core/test_circuit.py
MathieuG-Colibri: Removed the test to move it to the custom controlled gate branch for more clarity.

tests/core/instruction/gates/test_custom_gate.py
MathieuG-Colibri: Added the function https://github.com/ColibrITD-SAS/mpqp/pull/142/commits/ed43b21b5d3c902170d960dc17cc3fd00c4e164e

Henri-ColibrITD (approved):

hJaffaliColibritd (approved):

PR #147, opened 2025-05-20
Feat multiple observables

MathieuG-Colibri:
Added the handling of multiple observables by grouping Pauli monomials.


mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -998,6 +998,15 @@ def commutes_with(self, other: PauliString) -> bool:
             == 0
         )
 
+    def qubit_wise_commutes_with(self, other: PauliStringMonomial) -> bool:
```
hJaffaliColibritd: I would just modify `commutes_with` and add /use a parameter to select the method / definition of commutation.

You can reuse the Enum `CommutingTypes` you created in the `pauli_grouping.py`.

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -998,6 +998,15 @@ def commutes_with(self, other: PauliString) -> bool:
...
+        if len(other.atoms) != len(self.atoms):
```
hJaffaliColibritd: Use `nb_qubits` ?

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,143 @@
...
+
+class GroupingMethods(Enum):
+    GREEDY = auto()
+    COLORING_GREEDY = auto()
+    COLORING_SF = auto()
+    COLORING_LF = auto()
+    COLORING_RECURSIVE_LF = auto()
+    COLORING_DB = auto()
+    COLORING_DSATUR = auto()
+    CLIQUE_REMOVING = auto()
```
hJaffaliColibritd: This should be given as a parameter of `ExpectationMeasure`, where, if there is several observables, he can directly choose the grouping method when instantiating the measure.

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,143 @@
...
+
+    Returns:
+        A list of single qubit instructions.
+    """
+    result = []
+    for i, atoms in enumerate(group[0].atoms):
+        if atoms == X:
+            result.append(Ry(-np.pi / 2, i))
+        elif atoms == Y:
+            result.append(Rx(-np.pi / 2, i))
```
hJaffaliColibritd: This is already implemented in `PauliStringAtom` and stored in parameter `self._basis_change` which contains the list of NativeGates to add to the corresponding qubit. 
It avoids also the usage of Ry and Rx, and uses H and S_dagger.

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,143 @@
...
+            if type == CommutingTypes.QUBITWISE:
+                if all(monomial.qubit_wise_commutes_with(m_g) for m_g in group):
+                    group.append(monomial)
+                    added = True
+                    break
+            elif type == CommutingTypes.FULL:
+                if all(monomial.commutes_with(m_g) for m_g in group):
+                    group.append(monomial)
+                    added = True
+                    break
```
hJaffaliColibritd: This can be merged into a single call to `commutes_with` if you allow the commuting type as parameter

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,143 @@
...
-    pass
+def pauli_monomial_eigenvalues(monom: PauliStringMonomial) -> npt.NDArray[np.float64]:
+    result = np.array([1], dtype=np.float64)
+    eigen_I = np.array([1, 1])
+    eigen_XYZ = np.array([1, -1])
+    for atom in monom.atoms:
+        if atom == I:
+            result = np.kron(result, eigen_I)
+        else:
+            result = np.kron(result, eigen_XYZ)
```
hJaffaliColibritd: Same thing here, the eigenvalues corresponding to each `PauliStringAtom` are already accessible via `self._eig_vals`, and the same for eigenvectors if needed

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,143 @@
...
+    return result
 
 
-def qubit_wise_commutation_pauli_grouping(monomials: list[PauliStringMonomial]):
+def run_optimized_multi_observables(
```
hJaffaliColibritd: The `run_...` functions must be moved to `mpqp.execution.runner` file

And this function should automatically be called by `run` , and not directly used by the user. So you should integrate it in the existing execution workflow.

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,143 @@
...
+    circuit: QCircuit,
+    observable: Union[PauliString, Observable],
```
hJaffaliColibritd: you don't need a parameter observable, because the (and not only one) observables are already stored in `ExpectationMeasure` that is attached to the circuit normally.

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,143 @@
...
+    """
+    if isinstance(observable, Observable):
+        observable = observable.pauli_string
+    if grouping_method == GroupingMethods.GREEDY:
+        grouping = pauli_grouping_greedy(observable.monomials, commuting_type)
+    else:
+        raise ValueError("This type of grouping is not currently supported.")
+    result = 0
+    for group in grouping:
+        local_result = run(circuit + QCircuit(find_qubitwise_rotations(group)), device)
```
hJaffaliColibritd: I recommand the usage of `pre_measure` instead of adding artificially the gates like this. See other examples of usage of premeasure in BasisMeasure and ExpectationMeasure specific cases, handled in runner.py maybe.

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,143 @@
...
+        assert isinstance(local_result, Result)
+        for monom in group:
+            expectation_value = np.dot(
+                pauli_monomial_eigenvalues(monom), local_result.probabilities
+            )
+            result += expectation_value * monom.coef
+    return result
```
hJaffaliColibritd: How do you do when you have several observables ? 
Here you compute the probabilities while I think you should submit observable jobs to the device, especially that most of the time, this feature is thought for remote jobs. I think here the logic is not exactly respected.

tests/tools/test_pauli_grouping.py
```diff
@@ -0,0 +1,32 @@
...
+        (GOOGLEDevice.CIRQ_LOCAL_SIMULATOR),
+        (AWSDevice.BRAKET_LOCAL_SIMULATOR),
+    ],
+)
+def test_expectation_values_devices(device: AvailableDevice):
+    circuit = random_circuit(nb_qubits=3)
+    string = X @ I @ Z + X @ Z @ Z + I @ Z @ Z - 2 * Y @ Z @ Z + 3 * X @ Y @ Z
+    true_result: float = run(circuit + QCircuit([ExpectationMeasure(Observable(string))]), device).expectation_values  # type: ignore
+    decomposed_pauli = run_optimized_multi_observables(circuit, string, device)
+    assert round(true_result, 6) == round(decomposed_pauli, 6)
```
hJaffaliColibritd: Test should be done by calling directly `run` and have also remote tests with the --long-costly tag, especially for AWS

hJaffaliColibritd (changes requested):

tests/tools/test_pauli_grouping.py
```diff
@@ -0,0 +1,32 @@
...
+from mpqp.tools.circuit import random_circuit
+from mpqp.tools.pauli_grouping import run_optimized_multi_observables
+
+
+@pytest.mark.parametrize(
+    "device",
+    [
+        (IBMDevice.AER_SIMULATOR),
+        (GOOGLEDevice.CIRQ_LOCAL_SIMULATOR),
+        (AWSDevice.BRAKET_LOCAL_SIMULATOR),
```
hJaffaliColibritd, line 23: Pourquoi ne pas avoir mis ATOSDevice.MYQLM_PYLINALG ET CLINALG par exemple ?

hJaffaliColibritd (changes requested):

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: The problem with this is that we need to decompose the observable so it would have been extracted from the circuit.

That's why I've currently put it outside of the run function for testing purposes and also I would need to be a surgeon with the measurements.

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: My issue is that commutes_with is also a method of PauliString and qubit wise commutation isn't possible for Pauli strings...

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/552f1f6dfd677e30a1e41906f2d7ddc0f568c1b2

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/552f1f6dfd677e30a1e41906f2d7ddc0f568c1b2

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/552f1f6dfd677e30a1e41906f2d7ddc0f568c1b2

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/552f1f6dfd677e30a1e41906f2d7ddc0f568c1b2

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/552f1f6dfd677e30a1e41906f2d7ddc0f568c1b2

tests/tools/test_pauli_grouping.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/552f1f6dfd677e30a1e41906f2d7ddc0f568c1b2

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/552f1f6dfd677e30a1e41906f2d7ddc0f568c1b2

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/552f1f6dfd677e30a1e41906f2d7ddc0f568c1b2

but changed the basis of Y

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -455,26 +467,16 @@ def _check_targets_order(self):
...
-    ) -> list[list[PauliStringMonomial]]:
-        """Decompose the observables and regroup the pauli measurements
-        by commutativity relation using different strategies.
-
-        Args:
-            method: The grouping method to use.
-                - "full_clique": Finds the largest possible commuting groups (cliques).
-                - "full_greedy": Uses a greedy algorithm to iteratively build commuting groups.
-                - "qubit_wise_clique": Groups Pauli strings based on qubit-wise commutativity.
+    def get_pauli_grouping(self) -> list[list[PauliStringMonomial]]:
```
MoHermes, line 477: Doc?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -870,6 +887,10 @@ def positive_eigen_values(self) -> npt.NDArray[np.bool_]:
         eigvals = reduce(np.kron, [a.eigen_values for a in self.atoms])
         return eigvals > 0
 
+    @property
+    def name(self) -> str:
+        return f"{'@'.join(map(str,self.atoms))}"
```
MoHermes: spacing `map(str, self.atoms)`

mpqp/core/instruction/measurement/expectation_value.py
MathieuG-Colibri: Added https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/b424f663cb2887f0c211aa2c4bc6f52cbbf3d67e

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: Was fixed forgot in which commit though.

tests/tools/test_pauli_grouping.py
MathieuG-Colibri: The test was changed now it calls the same job with and without the Pauli grouping and compare the results.

tests/tools/test_pauli_grouping.py
MathieuG-Colibri: Note I have removed them because I was told to not do the my qlm implementation.

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: fixed

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -365,7 +369,10 @@ def __init__(
         observable: Union[Observable, list[Observable]],
         targets: Optional[list[int]] = None,
         shots: int = 0,
+        commuting_type: CommutingTypes = CommutingTypes.QUBITWISE,
```
Henri-ColibrITD, line 378: hmm, if this is used only in one (or two) method(s), I this it would make more sense to have these  as input to these methods I think, we can discuss this

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -455,26 +467,26 @@ def _check_targets_order(self):
...
-        Returns:
-            A list of list, where each list contains Pauli strings that can be measured simultaneously.
-        3M-TODO
+    def get_pauli_grouping(self) -> list[list[PauliStringMonomial]]:
+        """Return the grouped monomials of the Pauli string of the observable.
+        The grouping is done according to the grouping method of the expectation measure and the chosen commutativity type.
         """
-        ...
+        monomials = []
+        for obs in self.observables:
```
Henri-ColibrITD: I think I would need a quick explanation to understand the ins and outs of this method, core to the PR

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -840,7 +858,7 @@ class PauliStringMonomial(PauliString):
     """
 
     def __init__(self, coef: Coef = 1, atoms: Optional[list["PauliStringAtom"]] = None):
-        self.coef = coef
+        self.coef: Coef = coef
```
Henri-ColibrITD: needed for the doc ?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -870,6 +888,10 @@ def positive_eigen_values(self) -> npt.NDArray[np.bool_]:
         eigvals = reduce(np.kron, [a.eigen_values for a in self.atoms])
         return eigvals > 0
 
+    @property
```
Henri-ColibrITD, line 895: is this different from str ?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1211,7 +1244,12 @@ def __hash__(self):
     def to_matrix(self) -> npt.NDArray[np.complex64]:
         return self.matrix
 
-    def commutes_with(self, other: PauliString) -> bool:
+    def get_basis_change(self) -> list[type[SingleQubitGate]]:
```
Henri-ColibrITD, line 1258: this tells me that making the basis change a private variable is probably not necessary

mpqp/execution/providers/aws.py
```diff
@@ -122,6 +124,114 @@ def run_braket(job: Job, translation_warning: bool = True) -> Result:
     return extract_result(res, job, job.device)
 
 
+@typechecked
+def run_braket_observable(job: Job, translation_warning: bool = True):
```
Henri-ColibrITD, line 128: here too I would need a discussion to understand it properly

mpqp/execution/runner.py
```diff
@@ -87,6 +87,9 @@ def adjust_measure(measure: ExpectationMeasure, circuit: QCircuit):
         tweaked_observables,
         list(range(circuit.nb_qubits)),
         measure.shots,
```
Henri-ColibrITD, line 90: maybe we should pass the deep copied measure, instead of passing all the attributes individually

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,82 @@
...
+        type: the commuting type by which we want to group the Pauli operators.
 
     Returns:
         A list of list of Pauli monomials containing the fully commutativity grouping.
 
 
     Examples:
         >>> from mpqp.measures import I, X, Y, Z
-        >>> full_commutation_pauli_grouping_greedy([I@X@X, Y@Y@Z, I@I@I, -3*Z@Y@X, Y@X@Y, -Z@Z@Y, 2*X@X@Y])
+        >>> pauli_grouping_greedy([I@X@X, Y@Y@Z, I@I@I, -3*Z@Y@X, Y@X@Y, -Z@Z@Y, 2*X@X@Y],CommutingTypes.FULL )
```
Henri-ColibrITD:
```suggestion
        >>> pauli_grouping_greedy([I@X@X, Y@Y@Z, I@I@I, -3*Z@Y@X, Y@X@Y, -Z@Z@Y, 2*X@X@Y], CommutingTypes.FULL )
```

mpqp/tools/pauli_grouping.py
```diff
@@ -1,44 +1,82 @@
...
-    pass
+def pauli_monomial_eigenvalues(monom: PauliStringMonomial) -> npt.NDArray[np.float64]:
+    result = np.array([1], dtype=np.float64)
+    for atom in monom.atoms:
+        result = np.kron(result, atom.eigen_values)
+    return result
 
 
-def qubit_wise_commutation_pauli_grouping(monomials: list[PauliStringMonomial]):
+def full_commutation_pauli_grouping_ibm_clique(monomials: list[PauliStringMonomial]):
```
Henri-ColibrITD, line 83: was ist das ?

tests/test_doc.py
```diff
@@ -119,7 +119,10 @@
     rand_product_local_unitaries,
     rand_unitary_2x2_matrix,
 )
-from mpqp.tools.pauli_grouping import full_commutation_pauli_grouping_greedy
+from mpqp.tools.pauli_grouping import (
+    CommutingTypes,
+    pauli_grouping_greedy,
```
Henri-ColibrITD, line 133: I saw somewhere a Method enum, should we pass this enum to the function pauly_grouping and use greedy as argument instead of dedicating a function to the greedy method ?

tests/tools/test_pauli_grouping.py
```diff
@@ -0,0 +1,49 @@
...
+            run(
+                circuit + QCircuit([ExpectationMeasure(observable)]),
+                device,
+                translation_warning=False,
+            ).expectation_values  # type: ignore
+        )
+    equal = True
+    print(single_exp_values)
+    assert isinstance(true_result, dict)
+    for i in range(len(true_result)):
```
Henri-ColibrITD: maybe use a `all` instead of this for + equal var ?

Henri-ColibrITD (changes requested):
see comments

mpqp/core/instruction/measurement/expectation_value.py
MathieuG-Colibri: The issue with that is the fact that we need to know it before hand the only two places we can put them are with the run call or with ExpectationMeasure

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: Yes because I need to have the pauli atoms inside of the monomial without the coefficient.

It is important at the end of the process.

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: an old prototype when Hamza worked on it, it's because in the papers the grouping is done through a minimum clique algorithm on a graph.

The reason why it hasn't been implemented yet it's because we would need to add the library networkX to our dependencies so this implementation has been moved to another time.

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: No I've forgot to remove it but it was to try and remove some typing errors (didn't work anyway)

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -38,6 +40,22 @@
 Coef = Union[Real, float, Expr]
 
 
+class CommutingTypes(Enum):
```
hJaffaliColibritd, line 43: Docstring (see other Enums in mpqp to adopt the same documentation style)

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -38,6 +40,22 @@
...
+    FULL = auto()
+    QUBITWISE = auto()
+
+
+class GroupingMethods(Enum):
```
hJaffaliColibritd, line 50: Docstring (see other Enums in mpqp to adopt the same documentation style)

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -962,7 +984,9 @@ def __hash__(self):
         atoms_as_tuples = tuple((atom.label for atom in self.atoms))
         return hash(atoms_as_tuples)
 
-    def commutes_with(self, other: PauliString) -> bool:
+    def commutes_with(
+        self, other: PauliString, method: CommutingTypes = CommutingTypes.FULL
```
hJaffaliColibritd, line 992: `method` argument is not present in the docstring

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1211,7 +1244,12 @@ def __hash__(self):
     def to_matrix(self) -> npt.NDArray[np.complex64]:
         return self.matrix
 
-    def commutes_with(self, other: PauliString) -> bool:
+    def get_basis_change(self) -> list[type[SingleQubitGate]]:
+        return self._basis_change
+
+    def commutes_with(
+        self, other: PauliString, method: CommutingTypes = CommutingTypes.FULL
```
hJaffaliColibritd, line 1262: `method` argument is not present in the docstring

Also, doesn't it make sense to talk about commuting types for PauliStringAtoms? I suppose Full and Qubit-wise are the same in that case (and I also suppose there is no other definition for single qubit Pauli operators). In that case, we could override the method to remove the method argument

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -343,6 +346,8 @@ class ExpectationMeasure(Measure):
             will be applied.
         observable: Observable used for the measure.
         shots: Number of shots to be performed.
+        commuting_type: Used to group the observables when running the circuit.
```
hJaffaliColibritd:
```suggestion
        commuting_type: Type of commutation (grouped in an Enum) used for Pauli grouping when optimizing the observables measurement.
```

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -343,6 +346,8 @@ class ExpectationMeasure(Measure):
...
+        grouping_method: How the pauli groups are formed.
```
hJaffaliColibritd:
```suggestion
        grouping_method: Enum describing the method used for Pauli grouping.
```

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -365,7 +370,10 @@ def __init__(
         observable: Union[Observable, list[Observable]],
         targets: Optional[list[int]] = None,
         shots: int = 0,
+        commuting_type: CommutingTypes = CommutingTypes.QUBITWISE,
+        grouping_method: GroupingMethods = GroupingMethods.GREEDY,
         label: Optional[str] = None,
+        optimize_measurement: Optional[bool] = True,
```
hJaffaliColibritd, line 381: Absent in the docstring Args.

hJaffaliColibritd (changes requested):

tests/tools/test_pauli_grouping.py
hJaffaliColibritd: Yes, I agree, we can maybe comment them, and put a 3M-TODO for the future, so we don't forget to implement it. You can resolve this conversation when the TODO has been added

tests/test_doc.py
MathieuG-Colibri: I don't really understand the point being made, the enum is used as argument for the method and each methods get their own function to execute them.

I don't think we want all of the functions at the same place.

tests/tools/test_pauli_grouping.py
MathieuG-Colibri: MyQLM is on the way out so I feel like it's unlikely that we implement it one day.

mpqp/execution/runner.py
MathieuG-Colibri: We could pass a deepcopy and then replace the observable.s but this way seems more optimal.

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: I added an error if this method is called with anything else than full commutativity.

MathieuG-Colibri: [8a87e03](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/8a87e039254930494a659e21f2c80b92e657ef2d)

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: Done [8a87e03](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/8a87e039254930494a659e21f2c80b92e657ef2d)

tests/tools/test_pauli_grouping.py
MathieuG-Colibri: Done [8a87e03](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/8a87e039254930494a659e21f2c80b92e657ef2d)

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: In this case pre_measure is not really useful in the sense that the pre measure do not depends on the measurement but on the pauli group.

mpqp/core/instruction/measurement/pauli_string.py
Henri-ColibrITD: removed then ?

mpqp/execution/runner.py
Henri-ColibrITD: hmmmmmoké

mpqp/tools/pauli_grouping.py
Henri-ColibrITD: ok, and then why did we change just the name ? x)

tests/test_doc.py
Henri-ColibrITD: good point, I think I have a brain fart. This being said having a single big function would probably not be the end of the world

mpqp/core/instruction/measurement/expectation_value.py
Henri-ColibrITD: ok, I see your point, I agree with the choice done then

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: Yes

Henri-ColibrITD (approved):

tests/execution/test_multi_observable.py
```diff
@@ -3,14 +3,16 @@
 from mpqp import QCircuit
```
hJaffaliColibritd, line 4: I did some test running simple code on almost all devices. 

```python 
from mpqp.all import *
from mpqp.measures import X as pX, Y as pY
obs = Observable(np.diag([0.7, -1, 1, 1]))
obs2 = Observable( pX @ pY + pY @ pY + pX @ pX + pY@pX)
c = QCircuit([H(0), CNOT(0,1), Ry(0.5, 0), ExpectationMeasure([obs, obs2], shots=10000, optimize_measurement=True)])
r = run(c, [IBMDevice.AER_SIMULATOR, AWSDevice.BRAKET_LOCAL_SIMULATOR, GOOGLEDevice.CIRQ_LOCAL_SIMULATOR])
[print(result) for result in r]
```

image

I know we agreed that we don't overwork the myqlm part for the moment in terms of multi-observable measurement scheme, but at least what we can do is compute sequentially the expectation value for each observable to at least return a result that is coherent with the others (here 2 expectations values instead of only 1). What do you think ?

hJaffaliColibritd (changes requested):

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -38,6 +40,28 @@
...
+    """Enum regrouping the types of commutation supported by the PauliStrings."""
+
+    FULL = auto()
+    QUBITWISE = auto()
+
+
+class GroupingMethods(Enum):
+    """Enum regrouping the different algorithm used to organize all of
+    the groups of Pauli monomials.
+    """
```
hJaffaliColibritd:
```suggestion
    """Enum regrouping the different Pauli grouping algorithms."""
```

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -991,12 +1026,19 @@ def commutes_with(self, other: PauliString) -> bool:
...
-            sum(1 for a, b in zip(self.atoms, other.atoms) if not a.commutes_with(b))
-            % 2
-            == 0
-        )
+        if method == CommutingTypes.QUBITWISE:
+            return all(
+                self.atoms[i].commutes_with(other.atoms[i])
+                for i in range(self.nb_qubits)
+            )
+        else:
```
hJaffaliColibritd: is this for Full commutatitivity ?
if so maybe precise the CommutingType explicitly

hJaffaliColibritd (changes requested):

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1137,7 +1181,7 @@ def atoms(self) -> list["PauliStringAtom"]:
         return [self]
 
     @property
-    def coef(self):
+    def coef(self) -> Coef:  # type: ignore
```
JulienCalistoTD: use ` # pyright: ignore[...]`

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1238,6 +1289,10 @@ def commutes_with(self, other: PauliString) -> bool:
             raise ValueError(
                 f"Expected a PauliStringAtom in parameter but got {type(other).__name__}"
             )
+        if not method == CommutingTypes.FULL:
+            raise ValueError(
+                f"PauliStringAtoms can only fully commutes with each others, instead received {method}"
+            )
         return other == I or self == I or self == other
```
JulienCalistoTD:
```suggestion
        if method == CommutingTypes.FULL:
            return other == I or self == I or self == other
        raise ValueError(
            f"PauliStringAtoms can only fully commutes with each others, instead received {method}"
        )
```

mpqp/execution/providers/google.py
```diff
@@ -44,6 +48,242 @@ def run_google(job: Job, translation_warning: bool = True) -> Result:
...
+        RepetitionsStoppingCriteria,
+        measure_observables,
+    )
+    from mpqp.execution.job import JobStatus
+
+    variances = {}
+    if job.measure is None:
+        raise NotImplementedError("job.measure is None")
+    assert isinstance(job.measure, ExpectationMeasure)
+
```
JulienCalistoTD, line 81:
```suggestion

     # TODO: optimize
```

tests/execution/test_multi_observable.py
```diff
@@ -3,14 +3,16 @@
...
+from mpqp.execution.devices import AWSDevice, GOOGLEDevice
 from mpqp.execution.runner import _run_single  # pyright: ignore[reportPrivateUsage]
 from mpqp.gates import *
 
 
 def list_circuits():
     return [
         QCircuit([H(0), CNOT(0, 1)]),
-        QCircuit([H(0), X(1)]),
+        QCircuit([H(0), X(1)]),  # type: ignore
```
JulienCalistoTD: use ` # pyright: ignore[...]`

tests/execution/test_multi_observable.py
```diff
@@ -65,3 +67,78 @@ def test_sequential_versus_multi(
...
+        print(non_optimized)
+        print(optimized)
+        if isinstance(non_optimized.expectation_values, float) and isinstance(
+            optimized.expectation_values, float
+        ):
+            assert round(non_optimized.expectation_values, 10) == round(
+                optimized.expectation_values, 10
+            )
+        else:
+            assert all(round(non_optimized.expectation_values[f"observable_{i}"], 5) == round(optimized.expectation_values[f"observable_{i}"], 5) for i in range(len(non_optimized.expectation_values)))  # type: ignore
```
JulienCalistoTD, line 138: use ` # pyright: ignore[...]`

tests/tools/test_pauli_grouping.py
```diff
@@ -0,0 +1,46 @@
...
+    obs = [Observable(string), Observable(str2), Observable(str3)]
+    true_result = run(circuit + QCircuit([ExpectationMeasure(obs)]), device, translation_warning=False).expectation_values  # type: ignore
+    single_exp_values = []
+    for observable in obs:
+        single_exp_values.append(
+            run(
+                circuit + QCircuit([ExpectationMeasure(observable)]),
+                device,
+                translation_warning=False,
+            ).expectation_values  # type: ignore
```
JulienCalistoTD: use ` # pyright: ignore[...]`

tests/tools/test_pauli_grouping.py
```diff
@@ -0,0 +1,46 @@
...
+        (AWSDevice.BRAKET_LOCAL_SIMULATOR),
+    ],
+)
+def test_expectation_values_devices(device: AvailableDevice):
+    circuit = random_circuit(nb_qubits=3)
+    string = X @ I @ Z + X @ Z @ Z + I @ Z @ Z
+    str2 = I @ Z @ Z - 2 * Y @ Z @ Z + 3 * X @ Y @ Z
+    str3 = X @ X @ X + X @ I @ X + I @ X @ X
+    obs = [Observable(string), Observable(str2), Observable(str3)]
+    true_result = run(circuit + QCircuit([ExpectationMeasure(obs)]), device, translation_warning=False).expectation_values  # type: ignore
```
JulienCalistoTD: use ` # pyright: ignore[...]`

mpqp/execution/providers/aws.py
```diff
@@ -122,6 +124,123 @@ def run_braket(job: Job, translation_warning: bool = True) -> Result:
...
+        for group in grouping:
+            transpiled_pre_measure = QCircuit(
+                find_qubitwise_rotations(group)
+            ).to_other_language(Language.BRAKET, translation_warning)
+            job.status = JobStatus.RUNNING
+            if job.measure.shots == 0:
+                from copy import deepcopy
+
+                cirq = deepcopy(transpiled_circuit + transpiled_pre_measure)
+                cirq.state_vector()  # type: ignore
```
JulienCalistoTD: use ` # pyright: ignore[...]`

mpqp/execution/providers/aws.py
```diff
@@ -122,6 +124,123 @@ def run_braket(job: Job, translation_warning: bool = True) -> Result:
...
+    from braket.tasks import GateModelQuantumTaskResult
+
+    if job.circuit.transpiled_circuit is None:
+        transpiled_circuit = job.circuit.to_other_device(
+            job.device, translation_warning
+        )
+    else:
+        transpiled_circuit = job.circuit.transpiled_circuit
+    assert isinstance(transpiled_circuit, Circuit)
+    device = get_braket_device(job.device, is_noisy=bool(job.circuit.noises))  # type: ignore
```
JulienCalistoTD: use ` # pyright: ignore[...]`

tests/execution/test_multi_observable.py
```diff
@@ -65,3 +67,78 @@ def test_sequential_versus_multi(
     # TODO modify here to match the logic of dict and observable.label etc
     for r1, e2 in zip(seq_results, multi_result.expectation_values.values()):
         assert r1.expectation_values == e2
+
```
hJaffaliColibritd, line 70: Performed a simple test with this code and gave me wrong result for `GOOGLEDevice.CIRQ_LOCAL_SIMULATOR`

```python
from mpqp.all import *
from mpqp.measures import X as XX, I as II, Z as ZZ, Y as YY
o = Observable(XX @ YY + ZZ @ XX)
o2 = Observable(II @ XX + YY @ ZZ)
exp = ExpectationMeasure([o, o2], shots=0, optimize_measurement=True)
c = QCircuit([Rx(0.23, 0), Rz(24.23, 1), CNOT(0,1), exp])
print(run(c, [IBMDevice.AER_SIMULATOR, AWSDevice.BRAKET_LOCAL_SIMULATOR, ATOSDevice.MYQLM_PYLINALG, GOOGLEDevice.CIRQ_LOCAL_SIMULATOR]))
```

image

hJaffaliColibritd (changes requested):

tests/tools/test_pauli_grouping.py
MathieuG-Colibri: Done [66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

mpqp/execution/providers/aws.py
MathieuG-Colibri: Done [66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

mpqp/execution/providers/aws.py
MathieuG-Colibri: Done [66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

tests/execution/test_multi_observable.py
MathieuG-Colibri: Fixed ! [66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

MathieuG-Colibri: Added this code to the tests :)

tests/tools/test_pauli_grouping.py
MathieuG-Colibri: [66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

tests/execution/test_multi_observable.py
MathieuG-Colibri: [66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

tests/execution/test_multi_observable.py
MathieuG-Colibri: [66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: [66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: Done [66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

tests/execution/test_multi_observable.py
MathieuG-Colibri: Done a "dirty" implementation but it supports nicely multiple observables.
[66e9e9f](https://github.com/ColibrITD-SAS/mpqp/pull/147/commits/66e9e9f25dbe9a95c42978b7fe9de210f4f1e36b)

mpqp/tools/pauli_grouping.py
hJaffaliColibritd: By the decomposition of hte observable you mean decompositino in PauliString ? This is handled directly in the observable class.

Did this change with the refacto you did ? maybe we can discuss it in a call

mpqp/tools/pauli_grouping.py
hJaffaliColibritd: Yes, but it is still a change of basis that we add before measuring the qubits, and that one would need/want to remove without having to tweak or modify the circuit itself. That is why we introduced this pre-measure, which stands for any modification we do to the initial circuit happening at the end for measurement purposes, which is exactly corresponding to what we do for multi-observable thing

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: What I wanted to say is that commutation type doesn't make sense for PauliStringAtom. There is only one definition of commutation for single qubit Pauli Operator, meaning that FULLY = QUBIT_WISE in that case.

That is why I think we should remove the argument `method` , because there is no mathematical definition of commuting type for PauliStringAtom. It only starts to make sense from PauliStringMonomials (several qubits)

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: Yes that I know but afterwards we need to pick the pauli groups of the observable and run each group.

The issue is that the observable is in the Measure that already got removed from the circuit beforehand, it is simpler just passing the obs as param.

mpqp/core/instruction/measurement/pauli_string.py
MathieuG-Colibri: The issue is that if method is not put in the function is cannot overload PauliMonomial's method because it doesn't have the same signature.

mpqp/tools/pauli_grouping.py
hJaffaliColibritd: Ok, then if observables are no more in the circuit, let's keep it like this

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Ok, so maybe to not complicate too much for the moment, we keep the argument method, but we just don't use it at all, meaning we don't return any error neither, and we put a default value as you did.

You can resolve this comment when you did it

mpqp/tools/pauli_grouping.py
MathieuG-Colibri: I agree that it is pre measure but here the pre measure is done on each pauli groups so the pre_measure is not tied to the expectation measure.

Here our pre_measure is not suitable for this use case.

hJaffaliColibritd: Okep, let's keep it like this for the moment and we can come back to it later if needed

hJaffaliColibritd (approved):

PR #149, opened 2025-06-02
Fix aws remote

MoHermes:


mpqp/execution/connection/aws_connection.py
```diff
@@ -11,11 +11,26 @@
 from getpass import getpass
 from pathlib import Path
 
+import boto3
```
hJaffaliColibritd: Maybe import it where you need it.

hJaffaliColibritd (changes requested):

mpqp/execution/connection/aws_connection.py
MoHermes: here a3cc979

hJaffaliColibritd: Good I resolve

hJaffaliColibritd (approved):

mpqp/execution/connection/aws_connection.py
Henri-ColibrITD, line 1: I tested the AWS SSO, and here is my feedback:
- we have three levels: the `main menu`, the `AWS menu` and the `AWS prompts`. For now, if the credentials are correct, I have a message saying so and I get sent back on the `AWS menu`. This is unnecessary IMO: the the credentials are correct, the user would not need to re-enter new credentials, so you could (should) send them directly to the `main menu`.
- on wrong credentials, they should not be saved (they should not override the previous credentials, but even they should not be saved at all in my opinion)

MoHermes: Updated here b0dd5d7

Henri-ColibrITD (approved):

mpqp/execution/connection/aws_connection.py
```diff
@@ -1,4 +1,3 @@
-import os
```
hJaffaliColibritd, line 1: I tried the code and script to setup my SSO credentials.
I am not sure if I entered the right one or not (might have double clicked on paste), and I was waiting for the code to tell me.
It said first that SSO was successfully configured, but then printed a kind of error

image

hJaffaliColibritd (changes requested):

mpqp/execution/connection/aws_connection.py
hJaffaliColibritd: Also, after `` SSO authentification configured soccessfully`` , we come back to the menu of AWS , when we choose between IAM and SSO. I think when it is successful, we should directly come back to the main menu by pressing enter, or following the same logic as other providers

hJaffaliColibritd: WHen I try to get the recap, it tells me that there is an error getting AWS credentials.

image


When I look at .aws/credentials, they are updated correctly, but not the config file. So maybe the problem comes from here ?

hJaffaliColibritd: When I try to run some computations, not working: 
image

hJaffaliColibritd: I updated the `setup_connections` script by re-installing mpqp from this branch, but still an error with the credentials. I wonder if it doesn't come from AWS side ?

image

mpqp/execution/connection/aws_connection.py
```diff
@@ -40,37 +50,76 @@ def setup_aws_braket_account() -> tuple[str, list[Any]]:
...
             "An Amazon Braket account is already configured. Do you want to update it? [y/N] "
         )
         if decision.lower().strip() != "y":
             return "Canceled.", []
 
-    braket_auth_choices = QuestionNode(
-        "Choose your Amazon Braket authentication method: ",
-        [
-            AnswerNode("IAM (Identity and Access Management)", configure_account_iam),
-            AnswerNode("SSO (Single Sign-On)", configure_account_sso),
```
JulienCalistoTD, line 58: why did you removed `AnswerNode` ?

mpqp/execution/connection/aws_connection.py
```diff
@@ -40,37 +50,76 @@ def setup_aws_braket_account() -> tuple[str, list[Any]]:
...
+        )
+    elif method == "2":
+        msg, _ = configure_account_sso()
+        print(
+            colored(
+                f"[SSO Authentication] {msg}", "green" if "successful" in msg else "red"
+            )
+        )
+    else:
+        print(colored("Invalid selection.", "red"))
```
JulienCalistoTD, line 85: the message is printed but you come back instantly, so we don't see the error message.
maybe asked if we want to retry ?

mpqp/execution/connection/aws_connection.py
```diff
@@ -112,22 +164,29 @@ def update_aws_credentials_file(
...
-    print("Configuring IAM authentication for Amazon Braket...")
-    os.system("aws configure")
+    print("Please enter your IAM credentials for AWS:")
 
-    print("IAM authentication configured successfully.")
-    save_env_variable("BRAKET_AUTH_METHOD", "IAM")
-    save_env_variable("BRAKET_CONFIGURED", "True")
+    access_key_id = input("Enter AWS access key ID: ").strip()
+    secret_access_key = getpass("Enter AWS secret access key (hidden): ").strip()
+    region = input("Enter AWS region: ").strip()
```
JulienCalistoTD, line 173: can you put default value `us-east-1` ?

mpqp/execution/connection/aws_connection.py
hJaffaliColibritd: Solved by cleaning and remove all previous credentials

hJaffaliColibritd (approved):

mpqp/execution/connection/aws_connection.py
Henri-ColibrITD: because this is now done as terminal prompts and not in question tree anymore, because we asked that is we the credentials are correct we should go back directly to the root question node and apparently this was hard to do (I'm not 100% convinced but it's not a big deal)

mpqp/execution/connection/aws_connection.py
Henri-ColibrITD: should we ?

hJaffaliColibritd: I think we shouldn't

Henri-ColibrITD: we agreed that we should not because there are QPUs available in other regions too

mpqp/execution/connection/aws_connection.py
hJaffaliColibritd: Will be done in the future Trello task

PR #150, opened 2025-06-12
Feat non continuous targets

MathieuG-Colibri:
Adding non continuous and non ordered targets for custom gates.

### DIFF
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -263,7 +263,9 @@ class QCircuit:
         elif isinstance(components, CustomGate) and components.swaps != []:
             self.add(components.swaps)
             self.instructions.append(components)
-            self.add(components.swaps)
+            tmp = components.swaps.copy()
+            tmp.reverse()
+            self.add(tmp)
         else:
             self.instructions.append(components)
 
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -260,6 +260,10 @@ class QCircuit:
 
         if isinstance(components, NoiseModel):
             self.noises.append(components)
+        elif isinstance(components, CustomGate) and components.swaps != []:
+            self.add(components.swaps)
+            self.instructions.append(components)
+            self.add(components.swaps)
         else:
             self.instructions.append(components)
 
--- a/mpqp/core/instruction/gates/custom_gate.py
+++ b/mpqp/core/instruction/gates/custom_gate.py
@@ -4,7 +4,7 @@ cases, you can use :class:`mpqp.core.instruction.gates.custom_gate.CustomGate`
 to add your custom unitary operation to the circuit, which will be decomposed
 and executed transparently."""
 
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, Tuple
 
 from typeguard import typechecked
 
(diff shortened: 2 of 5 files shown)


mpqp/core/circuit.py
```diff
@@ -260,6 +260,12 @@ def add(self, components: OneOrMany[Instruction | NoiseModel]):
 
         if isinstance(components, NoiseModel):
             self.noises.append(components)
+        elif isinstance(components, CustomGate) and components.swaps != []:
+            self.add(components.swaps)
+            self.instructions.append(components)
+            tmp = components.swaps.copy()
+            tmp.reverse()
+            self.add(tmp)
```
Henri-ColibrITD, line 268: I think I would do this upon running, not adding.

Also, if you add the swaps, did you reorder the targets ?

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -163,6 +170,49 @@ def __repr__(self) -> str:
         label = f", \"{self.label}\"" if self.label else ""
         return f"CustomGate({UnitaryMatrix(self.matrix)}, {self.targets}{label})"
 
+    def __organize_non_continuous_targets(
+        self, targets: list[int], sorted: list[int]
+    ) -> Tuple[list[int], list[int]]:
```
Henri-ColibrITD, line 175: do you need a big T tuple ?

mpqp/core/circuit.py
MathieuG-Colibri: I feel like we need the swaps at this point to that it's compatible with to other language and so that the to_matrix is correct.

Also yes if the targets are not continuous then to_other_language will not like it so it is rearranged in the constructor.

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -163,6 +170,49 @@ def __repr__(self) -> str:
...
+            if sorted[i + 1] == sorted[i] + 1:
+                current += 1
+            else:
+                if current >= biggest_block:
+                    biggest_block = current
+                    start = i - biggest_block + 1
+                current = 1
+        if biggest_block < current:
+            biggest_block = current
+            start = len(sorted) - biggest_block
```
Henri-ColibrITD, line 192:
```suggestion
        shift = min(ts)
        shifted_ts = [t - shift for t in ts]
        swaps = {i + shift: t + shift for i, t in enumerate(shifted_ts) if i != t}
```

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -163,6 +170,49 @@ def __repr__(self) -> str:
...
+                    sorted[start + biggest_block + i],
+                    sorted[start + biggest_block - 1] + i + 1,
+                )
+            )
+            targets[targets.index(sorted[start + biggest_block + i])] = (
+                sorted[start + biggest_block - 1] + i + 1
+            )
+            sorted[start + biggest_block + i] = (
+                sorted[start + biggest_block - 1] + i + 1
+            )
```
Henri-ColibrITD, line 213:
```suggestion
        self.swaps = [SWAP(i, j) for i, j in swaps.items()]
```

mpqp/tools/maths.py
```diff
@@ -457,3 +457,69 @@ def is_power_of_two(n: int) -> bool:
...
     return n >= 1 and (n & (n - 1)) == 0
+
+
+@typechecked
+def successive_kron_products(matrices: list[Matrix]):
+    """This function performs several kronecker products in a row."""
+    result = matrices[0]
+    for matrix in matrices[1:]:
+        result = np.kron(result, matrix)
+    return result
```
Henri-ColibrITD, line 468: this can probably be replaced by
```py
reduce(np.kron, matrices)
```
wherever it is called

mpqp/tools/maths.py
```diff
@@ -457,3 +457,69 @@ def is_power_of_two(n: int) -> bool:
...
+
+    Returns:
+        The shuffled matrix according to the given targets.
+
+    Example:
+        >>> m1 = np.eye(2)
+        >>> m2 = np.array([[0,1], [1,0]])
+        >>> matrix = np.kron(m1,m2)
+        >>> matrix_swapped = np.kron(m2,m1)
+        >>> matrix_eq(matrix_swapped,rearrange_matrix(matrix, [1,0]))
```
Henri-ColibrITD, line 493: good for test but less pleasant for doc, I would just `pprint` the matrix

mpqp/tools/maths.py
```diff
@@ -457,3 +457,69 @@ def is_power_of_two(n: int) -> bool:
...
+        True
```
Henri-ColibrITD, line 494:
```suggestion
        >>> I = np.eye(2)
        >>> X = np.array([[0,1], [1,0]])
        >>> matrix = np.kron(I, X)
        >>> pprint(matrix)
        [[0, 1, 0, 0],
         [1, 0, 0, 0],
         [0, 0, 0, 1],
         [0, 0, 1, 0]]
        >>> pprint(rearrange_matrix(matrix, [1,0])))
        [[0, 0, 1, 0],
         [0, 0, 0, 1],
         [1, 0, 0, 0],
         [0, 1, 0, 0]]
```

mpqp/core/circuit.py
MoHermes, line 268: why not using ``reversed()`` instead of that reverse copy logic? so you eliminate the use of ``.copy()`` and ``.reverse()``.

Henri-ColibrITD:
we close this PR ? nothing else needs to be done ?

Henri-ColibrITD:
maybe we could keep the tests

PR #151, opened 2025-06-18
fix: string formatting in 'Result' for 'Google' devices

MoHermes:


hJaffaliColibritd (changes requested):
I think the changes in the ordering of the samples in the Result object, when running on GOOGLEDevice should be done in `google.py` when we extract the different results and instantiate the result. In that way we don't need to modify the class Result

mpqp/execution/providers/google.py
```diff
@@ -316,13 +316,12 @@ def extract_result_SAMPLE(
 
     data = [
         Sample(
-            bin_str="".join(map(bin, state)),
+            bin_str="".join(map(lambda s: format(s, f'0{nb_qubits}b'), state)),
```
hJaffaliColibritd: Elegant way of formatting.

hJaffaliColibritd (approved):

mpqp/execution/providers/google.py
Henri-ColibrITD: could be even better I think:
```suggestion
            bin_str="".join(map(lambda s: f'{s:0{nb_qubits}b}', state)),
```

mpqp/execution/result.py
```diff
@@ -487,6 +487,7 @@ def __str__(self):
                 f"Index: {sample.index}, Count: {sample.count}, Probability: {clean_number_repr(probability)}"
                 for sample, probability in zip(self.samples, probabilities)
             )
+
```
Henri-ColibrITD, line 490: 🤔

JulienCalistoTD (approved):

Henri-ColibrITD (approved):

PR #152, opened 2025-06-18
Feat cirq noise

MoHermes:


examples/notebooks/6_Noise_Simulation.ipynb
```diff
@@ -1133,7 +1221,7 @@
                     "text": [
                         "NoiseModel:\n",
                         "  Basis gates: ['cx', 'h', 'id', 'rz', 'sx', 'y']\n",
-                        "  Instructions with noise: ['y', 'cx', 'noisy_identity_0', 'noisy_identity_2', 'noisy_identity_1', 'h']\n",
+                        "  Instructions with noise: ['h', 'noisy_identity_1', 'y', 'noisy_identity_2', 'cx', 'noisy_identity_0']\n",
```
hJaffaliColibritd, line 1224: Why this changed ?

mpqp/core/circuit.py
```diff
@@ -1323,7 +1323,17 @@ def to_other_language(
                     cirq_instruction = instruction.to_other_language(Language.CIRQ)
                     cirq_circuit.append(cirq_instruction.on(*targets))
 
+            if self.noises:
```
hJaffaliColibritd, line 1402:
```suggestion
            if len(self.noises) != 0:
```

mpqp/execution/devices.py
```diff
@@ -423,6 +423,12 @@ def is_processor(self) -> bool:
         """
         return self.name.startswith("PROCESSOR")
 
+    def is_noisy_simulator(self) -> bool:
+        return self in [
+            GOOGLEDevice.CIRQ_LOCAL_SIMULATOR,
+            GOOGLEDevice.IONQ_SIMULATOR,
```
hJaffaliColibritd, line 429: Just by curiosity, did we have the opportunity (or tokens) to test the noisy simulation on IonQ simulator ? Are we sure it accepts all the noise models, or just the IonQ QPU noise ?

mpqp/execution/providers/google.py
```diff
@@ -10,6 +10,8 @@
     from cirq.study.result import Result as CirqResult
     from cirq.work.observable_measurement_data import ObservableMeasuredResult
 
+from cirq.circuits.circuit import Circuit
+from cirq.circuits.moment import Moment
```
hJaffaliColibritd: Move the imports to places only when it is needed. If you need for typing, add also the import in ``if TYPE_CHECKING:``.

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+@typechecked
+def apply_noise_to_cirq_circuit(
+    cirq_circuit: "Circuit",
+    noises: list[NoiseModel],
+    nb_qubits: int,
+) -> "Circuit":
+    """Apply noise models to a Cirq circuit.
+
+    This function applies noise models to a given Cirq circuit based on the specified noise models and
+    the number of qubits in the circuit. It modifies the original circuit by adding noise operations
```
hJaffaliColibritd: In practice, it does not modify the original circuit, but only builds a new circuit with noise operations from the original one. So better precising that it doesn't affect the original circuit

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+    after the original gates and returns a new circuit with the noise applied.
+
+    Args:
+        cirq_circuit: The Cirq circuit to apply noise to.
```
hJaffaliColibritd, line 43: I would say "The Cirq quantum circuit description on which we want to apply noise models" , to precise that we don't apply to this circuit, but only use it as a model

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+        noises: A list of noise models to apply to the circuit.
```
hJaffaliColibritd:
```suggestion
        noises: A list of NoiseModels to apply to the circuit.
```

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+        moment_ops = list(moment.operations)
+        noisy_moments.append(Moment(moment_ops))
+
+        qubit_noise_op: list[list[Operation]] = [[] for _ in range(nb_qubits)]
+
+        for op in moment_ops:
+            if isinstance(op.gate, (MeasurementGate, IdentityGate)):
+                continue
+
+            for noise in reversed(noises):
```
hJaffaliColibritd, line 83: I would reverse `noises` outsite the for loop of the moments and store it in a variable, so we don't reverse this list each time we have a new moment

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                cirq_noise = noise.to_other_language(Language.CIRQ)
```
hJaffaliColibritd: This also should be done only once at the beginning and stored, so we don't call `to_other_language` on eah nnoise model for each moment.

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                if TYPE_CHECKING:
+                    assert isinstance(cirq_noise, Gate)
+
+                noise_dimension = getattr(noise, "dimension", 1)
+
+                if not noise.targets or len(noise.targets) == nb_qubits:
```
hJaffaliColibritd:
```suggestion
                if len(noise.targets) == 0 or len(noise.targets) == nb_qubits:
```

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                    target_qubits = qubit_list
+                else:
+                    target_qubits = [qubit_list[i] for i in noise.targets]
+
+                allowed_gates = set()
+                if noise.gates:
```
hJaffaliColibritd:
```suggestion
                if len(noise.gates) != 0:
```

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                    for gate in noise.gates:
+                        gate_obj = (
+                            gate.cirq_gate if hasattr(gate, "cirq_gate") else gate
+                        )
```
hJaffaliColibritd: I am not fan of the `else gate`. 
Each NativeGate should have an attribute `cirq_gate`, so if the user has put a funky NativeGate in `noise.gates` and forgot to implement the property `cirq_gate` we shouldn't cover his back here, we should raise an error.

Also here putting `gate` will make gate_obj being either a cirq or mpqp gate, which is a bit weird

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                        gate_cls = (
+                            type(gate_obj)
+                            if not isinstance(gate_obj, type)
+                            else gate_obj
+                        )
+                        allowed_gates.add(gate_cls)
+
+                if noise_dimension == 2 and len(op.qubits) == 2:
+                    if all(q in target_qubits for q in op.qubits):
+                        if not allowed_gates or isinstance(
```
hJaffaliColibritd: `len(allowed_gates) == 0`

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                            op.gate, tuple(allowed_gates)
```
hJaffaliColibritd: Here you transform each time `allowed_gates` into a Tuple.

Can't we do it once for all when we finished to construct the `allowed_gates` set ?

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                        ):
+                            noisy_gate = cirq_noise.on(*op.qubits)
+                            noisy_moments.append(Moment([noisy_gate]))
+
+                elif noise_dimension == 1:
+                    for q in op.qubits:
+                        if q in target_qubits:
+                            if not allowed_gates or isinstance(
```
hJaffaliColibritd: same as above, in general avoid `if collection` to check that it is empty or not, and prefer the usage of `len`

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                                op.gate, tuple(allowed_gates)
+                            ):
+                                qubit_index = qubit_list.index(q)
+                                qubit_noise_op[qubit_index].append(cirq_noise.on(q))
+
+        while any(qubit_noise_op):
```
hJaffaliColibritd: Same remark as above, use `len` ?

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                        )
+                        allowed_gates.add(gate_cls)
+
+                if noise_dimension == 2 and len(op.qubits) == 2:
+                    if all(q in target_qubits for q in op.qubits):
+                        if not allowed_gates or isinstance(
+                            op.gate, tuple(allowed_gates)
+                        ):
+                            noisy_gate = cirq_noise.on(*op.qubits)
+                            noisy_moments.append(Moment([noisy_gate]))
```
hJaffaliColibritd: Here you create a single moment with a single gate in it ?
This can cause circuit space problem if you have another gate at the same time on the other qubits, no ?

mpqp/execution/result.py
```diff
@@ -836,7 +836,7 @@ def load_by_local_ids(result_ids: list[int]):
 
     def save(self):
         """Save a batch of results to the local storage and returns the
-        corresponding local ``id``s to be used when the results needs to be
+        corresponding local ``id``'s to be used when the results needs to be
```
hJaffaliColibritd:
```suggestion
        corresponding local ``id`` to be used when the results needs to be
```

mpqp/qasm/qasm_to_cirq.py
```diff
@@ -130,9 +130,9 @@ def __init__(
         def __repr__(self) -> str:
             return (
                 f'U('
-                f'theta={self.theta !r}, '
-                f'phi={self.phi!r}, '
-                f'lmda={self.lmda})'
+                f'{str(self.theta)[:5]}, '
+                f'{str(self.phi)[:5]}, '
+                f'{str(self.lmda)[:5]})'
```
hJaffaliColibritd: There are some rounding problems ? Why not using directly rounding or decimal truncating ?

hJaffaliColibritd (changes requested):
Good work in general, the noise application algorithm and trick was not easy to find. Some minor changes needed.

tests/execution/test_validity.py
```diff
@@ -508,14 +509,16 @@ def test_validity_noise_to_other_language(language: Language):
...
                 noise_build.to_other_language(language)
+
         elif language in [Language.MY_QLM] and not isinstance(
             noise_build, (Depolarizing, PhaseDamping)
         ):
             with pytest.raises(NotImplementedError):
                 noise_build.to_other_language(language)
+
         else:
             assert noise_build.to_other_language(language) is not None
```
hJaffaliColibritd, line 528: I add here an example that may be problematic : 

```python
from mpqp.all import *
from mpqp.noise import *
c = QCircuit([H(0), CNOT(0,1), H(1), CNOT(0,2), Z(1), BasisMeasure(shots=100000)])
c.add(Depolarizing(0.22))
print(run(c, [IBMDevice.AER_SIMULATOR, AWSDevice.BRAKET_LOCAL_SIMULATOR, GOOGLEDevice.CIRQ_LOCAL_SIMULATOR]))
```

It looks like this results in an infinite loop, because if I remove GoogleDevice, it works perfectly.

mpqp/execution/providers/google.py
hJaffaliColibritd: This maybe go into infinite loop `while True` , there is a problem here

hJaffaliColibritd (changes requested):

examples/notebooks/6_Noise_Simulation.ipynb
```diff
@@ -1248,44 +1378,53 @@
...
-                        " Probabilities: [0.24, 0.54, 0.13, 0.09]\n",
-                        " Samples:\n",
-                        "  State: 00, Index: 0, Count: 24, Probability: 0.24\n",
-                        "  State: 01, Index: 1, Count: 54, Probability: 0.54\n",
-                        "  State: 10, Index: 2, Count: 13, Probability: 0.13\n",
-                        "  State: 11, Index: 3, Count: 9, Probability: 0.09\n",
-                        " Error: None\n"
+                        "BatchResult: 3 results\n",
+                        "    Result: circuit 1, AWSDevice, BRAKET_LOCAL_SIMULATOR\n",
+                        "      Counts: [23, 58, 13, 6]\n",
```
Henri-ColibrITD, line 1390: 1st tab: 4 spaces, subsequent tabs: 2 spaces

mpqp/execution/providers/google.py
MoHermes: @hJaffaliColibritd your point is valid but here after modifying the logic so we're no longer using `InsertStrategy` from `cirq`, and current implementation adds each noisy gate in a separate moment for simplicity.

mpqp/execution/providers/google.py
MoHermes: I see your point Hamza, in this case, each pop(0) call reduces the size of those ` qubit_noise_op`, so the loop is guaranteed to terminate.

examples/notebooks/6_Noise_Simulation.ipynb
MoHermes: image

We agreed on this implementation last time @hJaffaliColibritd

examples/notebooks/6_Noise_Simulation.ipynb
MoHermes: I added a task on Trello to check that, it happens with Qiskit

mpqp/qasm/qasm_to_cirq.py
MoHermes: Here cf7f478

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+        for op in moment_ops:
+            if isinstance(op.gate, (MeasurementGate, IdentityGate)):
+                continue
+
+            for noise in reversed(noises):
+                cirq_noise = noise.to_other_language(Language.CIRQ)
+                if TYPE_CHECKING:
+                    assert isinstance(cirq_noise, Gate)
+
+                noise_dimension = getattr(noise, "dimension", 1)
```
JulienCalistoTD: why do you used `getattr` and not  `.dimension` ?

mpqp/core/circuit.py
MoHermes: Both have the same behaviour, but `if self.noises:` is easier and pythonic to check for an empty collection?

MoHermes: Ok discussed it with @Henri-ColibrITD, will go for the requested change.

mpqp/qasm/qasm_to_cirq.py
```diff
@@ -129,10 +129,7 @@ def __init__(
 
         def __repr__(self) -> str:
             return (
-                f'U('
-                f'theta={self.theta !r}, '
-                f'phi={self.phi!r}, '
-                f'lmda={self.lmda})'
+                f"U(" f"{self.theta:.3f}, " f"{self.theta:.3f}, " f"{self.theta:.3f})"
```
JulienCalistoTD:
```suggestion
                f"U({self.theta:.3f}, {self.theta:.3f}, {self.theta:.3f})"
```

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+    from cirq.ops.identity import IdentityGate
+    from cirq.ops.measurement_gate import MeasurementGate
+    from cirq.ops.raw_types import Gate, Operation
+
+    qubit_list = sorted(cirq_circuit.all_qubits())
+    noisy_moments = []
+
+    for moment in cirq_circuit:
+        moment_ops = list(moment.operations)
+        noisy_moments.append(Moment(moment_ops))
```
JulienCalistoTD, line 75: you rebuild `Moment` from `moment_ops` ? Can you not just put  `noisy_moments.append(moment)` ?

mpqp/execution/providers/google.py
Henri-ColibrITD: I disagree, we should keep the name in human format. If we chose to put the exact object then we should put the actual link to it, but since we already have the link in the type hint I find it redundant

Henri-ColibrITD: In general we try to avoid being too specific on the type in case we change implementation details (for instance if we decide to input `Sequence` instead of `list`, so in that case I would write the doc as `The noise models to apply to the circuit.`

mpqp/execution/providers/google.py
Henri-ColibrITD: a bit too verbose imo, if we specify it above already then we can leave this sentence as is

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+    """Apply noise models to a Cirq circuit.
+
+    This function applies noise models to a given Cirq circuit based on the specified noise models and
+    the number of qubits in the circuit. It modifies the original circuit by adding noise operations
+    after the original gates and returns a new circuit with the noise applied.
+
+    Args:
+        cirq_circuit: The Cirq circuit to apply noise to.
+        noises: A list of noise models to apply to the circuit.
+        nb_qubits: The number of qubits in the circuit.
```
Henri-ColibrITD: is this useful ? can't it be retrieved from the cirq object ?

mpqp/execution/devices.py
MoHermes: https://docs.ionq.com/guides/simulation-with-noise-models
Yes, I think we can run noisy simulations on IonQ’s simulator not sure, but have to test to confirm

mpqp/execution/providers/google.py
Henri-ColibrITD: since reversed returns an iterator, I'm not sure it really bring an efficiency loss

mpqp/execution/providers/google.py
Henri-ColibrITD: ok, here it's useful for sure, so if we have an init step, might as well store the revered iterator while we're at it

mpqp/execution/providers/google.py
Henri-ColibrITD: (we prefer avoiding the truthy and falsy values, so when we try to do is clearer when we read it, and because in some occasions you might get burned by truthy/falsy values, for instance let's say an api returns a string and if it crashed it returns None instead, if you test `if not api_return_value`, you would not differentiate from the case where the value is an empty string and the case the api actually crashed)

mpqp/execution/providers/google.py
Henri-ColibrITD: 3 nested ifs could be rewritten as a single one ?

mpqp/execution/providers/google.py
Henri-ColibrITD: same as above for nested ifs

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                            if not allowed_gates or isinstance(
+                                op.gate, tuple(allowed_gates)
+                            ):
+                                qubit_index = qubit_list.index(q)
+                                qubit_noise_op[qubit_index].append(cirq_noise.on(q))
+
+        while any(qubit_noise_op):
+            current_moment_ops = []
+            for qubit_ops in qubit_noise_op:
+                if qubit_ops:
```
Henri-ColibrITD: len?

mpqp/execution/providers/google.py
Henri-ColibrITD: unless a qubit_ops is falsy, in that case you'd never pop it and have an infinite loop ?

Henri-ColibrITD: Ah, no, I see the logic. It would be easier to review if the code was a tiny bit more explicit ^^"

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+                    current_moment_ops.append(qubit_ops.pop(0))
+            noisy_moments.append(Moment(current_moment_ops))
```
Henri-ColibrITD: ok, I understood the logic now, here is a proposition of make it easier (to me) to understand at a glance:
```suggestion
        moments = []
        for moment_index in max(len(ops) for ops in qubit_noise_op):
            moments.append([ops[moment_index] for ops in qubit_noise_op if len(ops) > moment_index])
```

Henri-ColibrITD: or even
```suggestion
        moments = [
            [ops[moment_index] for ops in qubit_noise_op if len(ops) > moment_index]
            for moment_index in max(len(ops) for ops in qubit_noise_op)
        ]
```

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,96 @@
...
+    from cirq.ops.raw_types import Gate, Operation
+
+    qubit_list = sorted(cirq_circuit.all_qubits())
+    noisy_moments = []
+
+    for moment in cirq_circuit:
+        moment_ops = list(moment.operations)
+        noisy_moments.append(Moment(moment_ops))
+
+        qubit_noise_op: list[list[Operation]] = [[] for _ in range(nb_qubits)]
```
Henri-ColibrITD: https://github.com/ColibrITD-SAS/mpqp/pull/152#discussion_r2208764995 now refers
to this line

mpqp/execution/providers/google.py
Henri-ColibrITD: in the line bellow, what does the all stand for, what for it mean for a Qid to
be truthy ?

mpqp/execution/providers/google.py
Henri-ColibrITD: so that if the attr doesn't exist it doesn't error ? these things happen when
the libs are poorly typed

mpqp/execution/providers/google.py
Henri-ColibrITD: nb: it's said in the return, maybe it's enough ?

mpqp/execution/providers/google.py
MoHermes: Check with new modify c6c8e85

tests/execution/test_validity.py
MoHermes: Test again with new modify c6c8e85

mpqp/execution/providers/google.py
MoHermes: And also rather than hardcoding checks for specific noise types

mpqp/execution/providers/google.py
MoHermes: we modify the circuit by inserting new noise operations after the original operations. We create a new list of moments that includes both the original operations and the noise, so we don't mutate the original moments.

But `noisy_moments.append(moment)` reusing the same immutable `Moment` object, and later when adding noise, the structure won't be correct.

mpqp/execution/providers/google.py
MoHermes: Here, I think Qid objects aren’t evaluated for truthiness, but it’s the boolean result of `q in target_qubits` that `all` checks

mpqp/execution/providers/google.py
MoHermes: c6c8e85

mpqp/execution/providers/google.py
MoHermes: c6c8e85

mpqp/execution/providers/google.py
MoHermes: c6c8e85

mpqp/core/circuit.py
MoHermes: c6c8e85

mpqp/execution/providers/google.py
MoHermes: Here 50a7425

mpqp/execution/providers/google.py
MoHermes: 50a7425

mpqp/execution/providers/google.py
MoHermes: Having `nb_qubits` is better for type-checking, as `qubits` return `set[Qid]` but I think it could be solved other way

mpqp/execution/providers/google.py
MoHermes: 813e992

mpqp/execution/providers/google.py
MoHermes: 813e992

mpqp/execution/result.py
MoHermes: 813e992

mpqp/execution/providers/google.py
MoHermes: 813e992

mpqp/execution/providers/google.py
Henri-ColibrITD: but then does it mean that `len(qubits) == nb_qubits`

MoHermes: yes yes, that comes with the last update you did, you want to modfiy that and we remove nb_qubits

mpqp/execution/providers/google.py
Henri-ColibrITD: i'd prefer we have a `DimensionalNoise` type, and check for this time, this would be more in line with the rest of the library

mpqp/noise/noise_model.py
```diff
@@ -690,15 +729,25 @@ def to_other_language(
 
                 return GeneralizedAmplitudeDamping(self.gamma, float(self.prob))
 
-        # TODO: MY_QLM implementation
```
Henri-ColibrITD, line 693: should this really be removed ?

mpqp/execution/providers/google.py
Henri-ColibrITD: i'll let you do it ?

mpqp/execution/providers/google.py
MoHermes: 5b7a419

mpqp/execution/providers/google.py
MoHermes: 5b7a419

mpqp/execution/providers/google.py
```diff
@@ -18,6 +20,104 @@
...
+
+        for op in moment_ops:
+            if isinstance(op.gate, (MeasurementGate, IdentityGate)):
+                continue
+
+            for noise in reversed(noises):
+                if isinstance(noise, DimensionalNoiseModel):
+                    noise_dimension = noise.dimension
+                else:
+                    noise_dimension = 1
```
Henri-ColibrITD:
```suggestion
                noise_dimension = noise.dimension if isinstance(noise, DimensionalNoiseModel) else 1
```

Henri-ColibrITD (approved):

examples/notebooks/6_Noise_Simulation.ipynb
hJaffaliColibritd: Ok we solve it in another branch

mpqp/execution/providers/google.py
hJaffaliColibritd: > I disagree, we should keep the name in human format. If we chose to put the exact object then we should put the actual link to it, but since we already have the link in the type hint I find it redundant

Ok I agree with you

tests/execution/test_validity.py
hJaffaliColibritd: Still not working on my side. You tested and it worked ?

MoHermes: image

MoHermes: But it takes long time than ususal, but when remove GoogleDevice, it's fast , but if you go for low shots, it's very fine

hJaffaliColibritd (approved):

PR #146, opened 2025-05-14
Feat controlled custom gate

MathieuG-Colibri:
Added the Custom Controlled Gate class with its to_other_language methods.

Also made CustomGate initialization more flexible for the user.

PR #153, opened 2025-07-02
fix: Handling of gates with parameters

LiRem101:
Created a fix for the handling of gates with parameters. If one compiled OpenQASM 2.0 to OpenQASM 3.0 and a custom gate in the original file took parameters, the compiling would crash if the gate was afterwards used in the OpenQASM 2.0 program. Likewise, the program would crash if a gate with parameters was called in an if-body.


Henri-ColibrITD:
Hi @LiRem101, nice to have a new collaborator! it seems that your fix is already incorporated in this PR: https://github.com/ColibrITD-SAS/mpqp/pull/146 

This being said, we don't have your tests so (if you wish) you can open a PR on the corresponding branch with the tests only. This way you'll appear as contributor of this PR

LiRem101:
Hi @Henri-ColibrITD , thanks for the welcome! :) I have completely overlooked the fix - but I'll add my tests as you suggested

PR #165, opened 2025-08-14
Feat to other language aws

JulienCalistoTD:


mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -123,6 +123,32 @@ def to_other_language(
...
+                    for elt in self.matrix.flatten()
+                    if isinstance(elt, Expr)
+                )
+            )
+
+            if len(gate_symbols) > 0:
+                if not printing:
+                    raise ValueError(
+                        "Custom gates defined with symbolic variables cannot be "
+                        "exported to Braket yet (Braket requires numeric matrices)."
```
Henri-ColibrITD:
```suggestion
                        "exported to Braket for now (only numerical matrices are supported)."
```

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -82,6 +83,25 @@ def _qiskit_parameter_adder(
     return qiskit_param
 
 
+@conditional_typechecked
+def _sympy_to_braket_param(val: Expr | float) -> "float | FreeParameter":
+    from sympy import Expr, Symbol
+    from braket.circuits import FreeParameter
+
+    if isinstance(val, Symbol):
+        return FreeParameter(str(val))
```
Henri-ColibrITD: I don't this this is needed, since if val is a `Symbol`, then it would also return itself from `free_symbols`. Unless this is a speed thing ?

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -308,6 +308,11 @@ def to_other_language(
 
             return QLMObservable(self.nb_qubits, matrix=self.matrix)
         elif language == Language.BRAKET:
+            # TODO: Braket do not handle pauli with coef because it use QASM2
```
Henri-ColibrITD:
```suggestion
            # TODO: Braket does not handle pauli with coef because it uses QASM2
```

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -308,6 +308,11 @@ def to_other_language(
...
+            #       We need to pass without coef and compute yourself
```
Henri-ColibrITD: weird mix of 1st person plural and 2nd person singular

mpqp/core/instruction/measurement/measure.py
```diff
@@ -64,8 +64,12 @@ def __init__(
...
-        in ascending order. This list of operations can for instance include swap gates
-        and U3 gates."""
+        """
+        List of gates inserted before the measurement, either to correctly
+        swap target qubits when required or to adjust the basis for a BasisMeasure.
+        These operations ensure that the only measurement performed at the end
+        of the circuit is a standard measurement in the computational basis,
+        with qubits arranged in ascending order. This list may, for example,
+        include SWAP gates and U3 gates.
+        """
```
Henri-ColibrITD: why this rewrite ? I don't find it much clearer, potentially even a bit wrong

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1098,10 +1099,7 @@ def commutes_with(
             )
         elif method == CommutingTypes.FULL:
             return (
-                sum(
-                    1 for a, b in zip(self.atoms, other.atoms) if not a.commutes_with(b)
-                )
-                % 2
+                sum(not a.commutes_with(b) for a, b in zip(self.atoms, other.atoms)) % 2
```
Henri-ColibrITD, line 1157: I didn't know that worked, noice

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1391,7 +1390,7 @@ def commutes_with(
                 f"Expected a PauliStringAtom in parameter but got {type(other).__name__}"
             )
         if method == CommutingTypes.FULL:
-            return other == pI or self == pI or self == other
+            return other.label == "I" or self.label == "I" or self.label == other.label
```
Henri-ColibrITD: why this change ?

mpqp/core/circuit.py
```diff
@@ -1369,16 +1369,33 @@ def to_other_language(
...
+                    continue
+                braket_instr = instruction.to_other_language(Language.BRAKET)
+                try:
+                    target = instruction.targets
+                    if isinstance(instruction, ControlledGate):
+                        target = instruction.controls + target
+                    braket_circuit.add_instruction(braket_instr, target=target)
+                except:
+                    print(braket_instr)
+                    print(type(braket_instr))
```
Henri-ColibrITD: dunno about printing here but not a big deal

mpqp/execution/providers/aws.py
```diff
@@ -225,11 +238,16 @@ def run_braket_observable(job: Job):
...
+            from braket.circuits.observables import Sum
 
             copy = deepcopy(transpiled_circuit)
             braket_obs = obs.to_other_language(Language.BRAKET)
+            if isinstance(braket_obs, Sum):
+                targets = [job.measure.targets] * len(braket_obs.summands)
+            else:
+                targets = job.measure.targets
             copy.expectation(  # pyright: ignore[reportAttributeAccessIssue]
-                observable=braket_obs, target=job.measure.targets
```
Henri-ColibrITD, line 228: this looks a bit strange to me, we could discuss this at some point

mpqp/tools/pauli_grouping.py
```diff
@@ -55,6 +55,7 @@ def pauli_grouping_greedy(
         [[pI@pX@pX, pY@pY@pZ, pI@pI@pI], [-3*pZ@pY@pX, -1*pZ@pZ@pY], [pY@pX@pY], [2*pX@pX@pY]]
     """
     groups: list[list[PauliStringMonomial]] = []
+
```
Henri-ColibrITD, line 58: Image

tests/core/instruction/gates/test_custom_gate.py
```diff
@@ -53,12 +52,7 @@ def test_random_orthogonal_matrix(circ_size: int, device: AvailableDevice):
     for _ in range(targets_start + gate_size, circ_size):
         exp_state_vector = np.kron(exp_state_vector, np.array([1, 0]))
 
-    with (
-        pytest.warns(UnsupportedBraketFeaturesWarning)
```
Henri-ColibrITD, line 50: U happy ? x)

tests/core/instruction/gates/test_custom_gate.py
```diff
@@ -186,13 +165,11 @@ def _test_execution_equivalence(
...
     targets = [position for _, position in gates_n_positions]
 
-    with (
-        pytest.warns(UnsupportedBraketFeaturesWarning)
-        if isinstance(device, AWSDevice)
-        else contextlib.suppress()
-    ):
-        result_custom_gate = run(QCircuit([CustomGate(matrix, targets)]), device)
-        result_circuit = run(circuit, device)
+    print(matrix)
```
Henri-ColibrITD: not a fan of the prints, let people add the needed prints if they want to, and clean it up once the test is fixed

Henri-ColibrITD (commented):
I overall like the PR, small tweaks but it will be fast

mpqp/core/instruction/measurement/pauli_string.py
Henri-ColibrITD:
```suggestion
            return other is pI or self is pI or self is other
```

tests/core/instruction/gates/test_custom_gate.py
JulienCalistoTD: VERY !

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: removed: [4fbbf54](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/4fbbf5478f830810ed88dcb69729c2a3d9bfacd4)

mpqp/core/instruction/measurement/expectation_value.py
JulienCalistoTD: Renamed: [4fbbf54](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/4fbbf5478f830810ed88dcb69729c2a3d9bfacd4)

mpqp/core/instruction/measurement/measure.py
JulienCalistoTD: rebeased: [4fbbf54](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/4fbbf5478f830810ed88dcb69729c2a3d9bfacd4)

mpqp/core/circuit.py
JulienCalistoTD: changed the raise message : [4fbbf54](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/4fbbf5478f830810ed88dcb69729c2a3d9bfacd4)

tests/core/instruction/gates/test_custom_gate.py
JulienCalistoTD: removed: [4fbbf54](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/4fbbf5478f830810ed88dcb69729c2a3d9bfacd4)

mpqp/core/instruction/measurement/measure.py
Henri-ColibrITD: wrong commit hash I think

Henri-ColibrITD (approved):

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -346,7 +364,14 @@ def to_other_language(
...
             return self.qiskit_gate()
         elif language == Language.BRAKET:
-            return self.braket_gate()
+            from braket.circuits import Instruction
+
+            connection = self.targets
+            if isinstance(self, ControlledGate):
+                connection += self.controls
+            return Instruction(
+                operator=self.braket_gate(), target=list(range(len(connection)))
```
MoHermes: ` target=list(range(len(connection))` we don't end up here applying on the wrong qubits?
I think `target = connection` preserves the correct qubit mapping?

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -380,7 +384,7 @@ def __init__(
         targets: Optional[list[int]] = None,
         shots: int = 0,
         commuting_type: CommutingTypes = CommutingTypes.QUBITWISE,
-        grouping_method: GroupingMethods = GroupingMethods.GREEDY,
+        grouping_method: GroupingMethods = GroupingMethods.QISKIT,
```
MoHermes: Here, just to reduce the no. of measurements?

mpqp/core/circuit.py
```diff
@@ -1383,16 +1383,33 @@ def to_other_language(
...
+            for instruction in circuit.instructions:
+                if isinstance(instruction, (Barrier, Breakpoint)):
+                    continue
+                if isinstance(instruction, Measure):
+                    if not skip_pre_measure:
+                        for pre_measure in instruction.pre_measure:
+                            bracket_pre_measure = pre_measure.to_other_language(
+                                Language.BRAKET
+                            )
+                            braket_circuit.add(bracket_pre_measure, instruction.targets)
```
hJaffaliColibritd, line 1407: There may be a problem here, since the measure in current MPQP circuits can be put anywhere in the `instructions` list (even if the measure happens at the end of the circuit), you can end up with pre-measure instructions at the beginning/middle of the `braket_circuit` instead of the end (because the pre_measure is meant to come at the end, after all gates).
So I would find a way to add this pre_measure to braket circuit only after having translated the rest of the circuit

mpqp/core/circuit.py
```diff
@@ -1383,16 +1383,33 @@ def to_other_language(
...
+                    if not skip_measurements:
+                        if isinstance(instruction, BasisMeasure):
+                            braket_circuit.measure(instruction.targets)
```
hJaffaliColibritd, line 1410: Same remark as above, if the measure is not placed at the end of MPQP, it still has to be placed at the end of the braket circuit, which here will not be the case

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -121,6 +121,32 @@ def to_other_language(
...
+                        "exported to Braket for now (only numerical matrices are supported)."
+                    )
+                return None
+            else:
+                from braket.circuits import Instruction as BraketInstruction
+                from braket.circuits.gates import Unitary as BraketUnitary
+
+                return BraketInstruction(
+                    operator=BraketUnitary(self.definition.matrix),
+                    target=list(range(self.nb_qubits)),
```
hJaffaliColibritd: Shouldn't we use the targets of the custom gate ? Because the targets in a CustomGate require that they are ordered and contiguous, but not that they necessarly start from 0 and span the whole circuit.
So if the targets of the CustomGate are [2,3,4], this line will cause a problem

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -224,24 +241,25 @@ def to_other_language(
...
-                    "Symbolic expressions are not yet supported for braket "
-                    "export, this feature is coming very soon!"
-                )
-            return self.braket_gate(theta)
+            connection = self.targets
+            if isinstance(self, ControlledGate):
+                connection += self.controls
+            return Instruction(
+                operator=self.braket_gate(_sympy_to_braket_param(theta)),
+                target=list(range(len(connection))),
```
hJaffaliColibritd: Same problem here, the target argument should correspond to the gate's targets (here `connection`) and not just to the list of indices from 0 to `len(connection)-1`

mpqp/core/instruction/gates/native_gates.py
hJaffaliColibritd: I agree

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -430,7 +455,14 @@ def to_other_language(
...
             return self.qiskit_gate()
         elif language == Language.BRAKET:
-            return self.braket_gate()
+            from braket.circuits import Instruction
+
+            connection = self.targets
+            if isinstance(self, ControlledGate):
+                connection += self.controls
+            return Instruction(
+                operator=self.braket_gate(), target=list(range(len(connection)))
```
hJaffaliColibritd: Identity gate is not a ControlledGate so no need to check that and to include the self.controls, and again here the target of the Braket Instruction is wrong (just and int, and not always 0)

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -1079,20 +1111,19 @@ def to_other_language(
...
-                raise NotImplementedError(
-                    "Symbolic expressions are not yet supported for braket "
-                    "export, this feature is coming very soon!"
-                )
+            from braket.circuits import Instruction
 
-            return self.braket_gate(self.theta, self.phi, self.gamma)
+            connection = self.targets
+            if isinstance(self, ControlledGate):
+                connection += self.controls
```
hJaffaliColibritd: `U` can't be a controlled gate since it is a one qubit gate, so no need for this part

mpqp/core/instruction/gates/native_gates.py
```diff
@@ -1079,20 +1111,19 @@ def to_other_language(
...
+            return Instruction(
+                operator=self.braket_gate(
+                    _sympy_to_braket_param(self.theta),
+                    _sympy_to_braket_param(self.phi),
+                    _sympy_to_braket_param(self.gamma),
+                ),
+                target=list(range(len(connection))),
```
hJaffaliColibritd: Target here is just an int, and it's wrong

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -80,8 +80,8 @@ def __init__(
         self._is_diagonal = None
         self._diag_elements: Optional[npt.NDArray[np.float64]] = None
         self.label = label
-        self.transpile = None
         "See parameter description."
+        self.pre_transpile = None
```
hJaffaliColibritd: I would rename it `pre_transpiled`. And provide a description for this attribute.
Why do we need to pre-transpile the measure ? Isn't it  just sufficient to pre transpile the observables?
Or it has to do with the Pauli grouping part (in that case I would rename the attribute to a find a more explicit name)?

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -306,6 +306,10 @@ def to_other_language(
 
             return QLMObservable(self.nb_qubits, matrix=self.matrix)
         elif language == Language.BRAKET:
+            # TODO: Braket does not handle pauli with coef because it uses QASM2
+            # if self._pauli_string:
+            #     return self.pauli_string.to_other_language(Language.BRAKET)
+            # else:
```
hJaffaliColibritd: Apparently there `Sum` which is a way to define PauliString in Braket, which is linear combination of braket `Observable`s, which is something you used also in `pauli_string.py` line 833.

https://amazon-braket-sdk-python.readthedocs.io/en/latest/_apidoc/braket.circuits.observable.html#braket.circuits.observable.Observable.Sum

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -397,6 +401,7 @@ def __init__(
         """See parameter description."""
         self.optimize_measurement = optimize_measurement
         """See parameter description."""
+        self.pre_transpile = None
```
hJaffaliColibritd: I would rename it `pre_transpiled`. And provide a description for this attribute.
Why do we need to pre-transpile the measure ? Isn't it  just sufficient to pre transpile the observables?
Or it has to do with the Pauli grouping part (in that case I would rename the attribute to a find a more explicit name)?

mpqp/core/instruction/measurement/expectation_value.py
hJaffaliColibritd: I would rename it `pre_transpiled`. And provide a description for this attribute.

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -500,6 +506,29 @@ def get_pauli_grouping(self) -> list[list[PauliStringMonomial]]:
...
+            from qiskit.quantum_info import PauliList
+
+            pauli_labels = [mono.short_name for mono in unique_monos]
+            pauli_list = PauliList(pauli_labels)
+
+            # Choose grouping based on commutativity type
+            if self.commuting_type == CommutingTypes.QUBITWISE:
+                grouped = pauli_list.group_qubit_wise_commuting()
+            else:
+                grouped = pauli_list.group_commuting()
```
hJaffaliColibritd: if would put a elif with the right CommutingType instead of and else, and raise an error for else

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -500,6 +506,29 @@ def get_pauli_grouping(self) -> list[list[PauliStringMonomial]]:
...
+
+            grouped_monomials = [
+                [
+                    PauliString.from_str(
```
hJaffaliColibritd, line 563: Would it be faster if we have the same method for PauliStringMonomials?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -57,6 +57,7 @@ class GroupingMethods(Enum):
     COLORING_DB = auto()
     COLORING_DSATUR = auto()
     CLIQUE_REMOVING = auto()
+    QISKIT = auto()
```
hJaffaliColibritd: not sure it makes sense because Qiskit has several grouping methods also. That is why we gave proper names,  and not just the default qiskit one

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -820,15 +822,15 @@ def to_other_language(
...
                 braket_mono = mono.to_other_language(Language.BRAKET)
-                pauli_string = (
-                    pauli_string + braket_mono
-                    if pauli_string is not None
-                    else braket_mono
-                )
-            return pauli_string
+                pauli_string.append(braket_mono)
+
+            return Sum(pauli_string)
```
hJaffaliColibritd: `return Sum([mono.to_other_language(Language.BRAKET) for mono in self.monomials])`

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -963,6 +965,10 @@ def __deepcopy__(self, memo: Optional[dict[int, Any]] = None):
     def name(self) -> str:
         return f"{'@'.join(map(str, self.atoms))}"
 
+    @property
+    def short_name(self) -> str:
```
hJaffaliColibritd, line 1001: Doc ? Because it may not be directly clear what is the difference between name and short name (maybe because `name` was originally not the best choice).
Why do we need it to be a property?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1182,8 +1185,15 @@ def to_other_language(
...
-            return self.coef * reduce(matmul, braket_atoms)
+            from braket.circuits.observables import TensorProduct
+
+            if len(braket_atoms) == 1:
+                return (
+                    self.coef * braket_atoms[0]
+                )  # pyright: ignore[reportOperatorIssue]
+            return self.coef * TensorProduct(
+                braket_atoms
+            )  # pyright: ignore[reportOperatorIssue]
```
hJaffaliColibritd: If I am not mistaken, in braket we can avoid identity atoms when we defined monomials, and just put the pauli operators acting non-trivially on the qubits.
Do you think it's something we can use to avoid having big tensor products? (I don't know if Brakte simplfies behind the monomial)

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1182,8 +1185,15 @@ def to_other_language(
                 atom.to_other_language(Language.BRAKET)
                 for atom in self.atoms  # pyright: ignore[reportAssignmentType]
             ]
-
-            return self.coef * reduce(matmul, braket_atoms)
+            from braket.circuits.observables import TensorProduct
+
+            if len(braket_atoms) == 1:
```
hJaffaliColibritd: It's equivalent to checking the number of qubits of the monomial right? (thus can be done before generating  the list of braket atoms)

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1407,7 +1417,7 @@ def commutes_with(
                 f"Expected a PauliStringAtom in parameter but got {type(other).__name__}"
             )
         if method == CommutingTypes.FULL:
-            return other == pI or self == pI or self == other
+            return other is pI or self is pI or self is other
```
hJaffaliColibritd, line 1493: Why do we replace `==` by `is`?
In this case it's equivalent right? Because only one unique instance of each Pauli atom is allowed

mpqp/execution/providers/atos.py
```diff
@@ -252,10 +252,10 @@ def generate_observable_job(myqlm_circuit: "Circuit", job: Job) -> list["JobQLM"
         assert job.measure is not None and isinstance(job.measure, ExpectationMeasure)
     result = []
     for obs in job.measure.observables:
-        if obs.transpile is None:
+        if obs.pre_transpile is None:
             qlm_obs = obs.to_other_language(Language.MY_QLM)
         else:
-            qlm_obs = obs.transpile
+            qlm_obs = obs.pre_transpile
```
hJaffaliColibritd: (Maybe done in the vqa remote branch already)
We need that obs.pre_transpile is a dictionnary indexed by device/provider, otherwise an already transpiled observable in another language will end up in your `qlm_obs` variable.

mpqp/execution/providers/aws.py
```diff
@@ -154,25 +155,39 @@ def run_braket_observable(job: Job):
...
+
+    results, errors = {}, {}
     if job.measure.optimize_measurement:
-        grouping = job.measure.get_pauli_grouping()
         from mpqp.tools.pauli_grouping import (
             find_qubitwise_rotations,
             pauli_monomial_eigenvalues,
         )
 
+        if job.measure.pre_transpile is None:
```
hJaffaliColibritd: We don't fill the variable `job.measure.pre_transpile` after computation, is it in purpose that it have to be recomputed each time?
This variable should also be a dictionnary indexed by provider/device.

mpqp/execution/providers/aws.py
```diff
@@ -294,31 +321,52 @@ def submit_job_braket(job: Job) -> tuple[str, "QuantumTask"]:
...
+        # rebind safe_retrieve_samples from braket to Normalize the probability
+        # because the bracket does not do so and this causes a crash.
+        from braket.default_simulator.state_vector_simulation import (
+            StateVectorSimulation,
+        )
+
+        def safe_retrieve_samples(self):  # pyright: ignore[reportMissingParameterType]
+            probs = self.probabilities
+            probs = probs / np.sum(probs)
+            return np.random.choice(len(self._state_vector), p=probs, size=self._shots)
```
hJaffaliColibritd, line 372: What is this wizardery ?

Retrieving the probabilities is already handled by the class `Result` of mpqp when you give only the complex amplitudes of the quantum states, it's not sufficient to take care of this normalization problem?

mpqp/execution/runner.py
```diff
@@ -206,6 +206,7 @@ def _run_single(
     device: AvailableDevice,
     values: "Optional[dict[Expr | str, Complex]]" = None,
     display_breakpoints: bool = True,
+    reservation_arn: Optional[str] = None,
```
hJaffaliColibritd: This comes from the work done on the AWS reservation arn, but as agreed with @MoHermes something more generic should be passed to handle provider's specific run options (and not adding each time a specific argument).
This work is being updated in the VQA remote branch, maybe before merging to dev this version we need to finalize the VQA remote branch ? or merge this one in the VQA remote and then merge the big one to dev ?

To be discussed, the two branches becoming a bit too much interdependent from what I see

hJaffaliColibritd (changes requested):
Good work overall bravo! I would add more tests for to_other_language, to test extensively the new translation way

mpqp/core/instruction/measurement/expectation_value.py
JulienCalistoTD: Not the nº of measurements but the speed to calulate the grouping

mpqp/core/circuit.py
JulienCalistoTD: If you add a gate after a measurement, braket will raise an error.

mpqp/core/circuit.py
JulienCalistoTD: same, braket will raise an error.

mpqp/core/instruction/measurement/pauli_string.py
JulienCalistoTD: yes and it is faster with `is`

mpqp/execution/providers/aws.py
JulienCalistoTD: rebind safe_retrieve_samples from braket to Normalize the probability, because the bracket does not do so and this causes a crash.

mpqp/execution/runner.py
JulienCalistoTD: this was needed for HDES, i removed it in [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)

mpqp/execution/providers/aws.py
JulienCalistoTD: not in this branch, we need to rethink it in VQA

mpqp/execution/providers/atos.py
JulienCalistoTD: yes, in the vqa branch

mpqp/core/instruction/measurement/pauli_string.py
JulienCalistoTD: Yes change to self.nb_qubits [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)

mpqp/core/instruction/measurement/pauli_string.py
JulienCalistoTD: Yes, I changed it, but it's a bit annoying to remove all the identities.

mpqp/core/instruction/measurement/pauli_string.py
JulienCalistoTD: done : [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)
We need it to create qiskit pauli.

mpqp/core/instruction/measurement/pauli_string.py
JulienCalistoTD: Yes that's better ! : [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)

mpqp/core/instruction/measurement/expectation_value.py
JulienCalistoTD: Not faster

mpqp/core/instruction/measurement/expectation_value.py
JulienCalistoTD: Done: [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)

mpqp/core/instruction/measurement/expectation_value.py
JulienCalistoTD: Done: [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492) 
We need to reviewed pre_transpiled for vqa, so description could wait

mpqp/core/instruction/measurement/expectation_value.py
JulienCalistoTD: Done: [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)
We need to reviewed pre_transpiled in vqa branch

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: Done: [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: Done: [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: Done: [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: list(range(3)) = [0, 1, 2]
Done : [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)

mpqp/core/instruction/gates/custom_gate.py
JulienCalistoTD: Done: [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)
But in fact, we don't need the actual target for this translation, because the target is handled when we add it to the circuit.

mpqp/core/instruction/gates/native_gates.py
JulienCalistoTD: Done: [330caa0](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/330caa0640821748212944e7c1f7913cc2a29492)

mpqp/core/instruction/measurement/expectation_value.py
JulienCalistoTD: No it was because we pass through qasm2 and qasm_to_braket was not handle the coef. But was now we translate directly to braket we can
done: [b5c7ac2](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/b5c7ac2bf78ca76dd061eac45017a053a93a3377)

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -961,8 +977,14 @@ def __deepcopy__(self, memo: Optional[dict[int, Any]] = None):
 
     @property
     def name(self) -> str:
+        """Return a monomial without the coefficient."""
         return f"{'@'.join(map(str, self.atoms))}"
 
+    @property
+    def short_name(self) -> str:
+        """Return a monomial without the coefficient and tensor product."""
```
hJaffaliColibritd:
```suggestion
        """Returns the string associated to the monomial without the coefficient and tensor product symbol."""
```

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -961,8 +977,14 @@ def __deepcopy__(self, memo: Optional[dict[int, Any]] = None):
 
     @property
     def name(self) -> str:
+        """Return a monomial without the coefficient."""
```
hJaffaliColibritd:
```suggestion
        """Returns the string associated to the monomial without the coefficient."""
```

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Oh interesting

mpqp/core/instruction/measurement/pauli_string.py
JulienCalistoTD: GroupingMethods is from `QISKIT`
but CommutingTypes can change:

`mpqp\core\instruction\measurement\expectation_value.py`    -> L. 545
```py
      elif self.grouping_method == GroupingMethods.QISKIT:
            from qiskit.quantum_info import PauliList

            pauli_labels = [mono.short_name for mono in unique_monos]
            pauli_list = PauliList(pauli_labels)

            # Choose grouping based on commutativity type
            if self.commuting_type == CommutingTypes.QUBITWISE:
                grouped = pauli_list.group_qubit_wise_commuting()
            elif self.commuting_type == CommutingTypes.FULL:
                grouped = pauli_list.group_commuting()
            else:
                raise NotImplementedError(
                    f"{self.commuting_type} is not yet supported."
                )
  ```

hJaffaliColibritd: GroupingMethods is an object that is defined in MPQP. In Qiskit documentation, you will see that there different grouping methods available. The idea was to have different grouping methods regrouped in the mpqp.tools on which we have control, and if we use the qiskit one, it will be there, but at least we know which algorithm is running for grouping.
Saying just Qiskit is unclear, because it does not tell us which algo is running behind, and if they change the default algo, it changes each time.

mpqp/core/circuit.py
hJaffaliColibritd: Ok but here pre-measure are gates, so it will not be detected by braket, and they need to be at the end for sure, (which is not ensured here). I would put them at the end manually before the measure.

mpqp/core/circuit.py
hJaffaliColibritd: What about the error to catch the error in a proper way? Or the error message is clear enough?

mpqp/execution/providers/aws.py
hJaffaliColibritd: Oh ok I see, so you redefine a function that is used in Braket. I resolve

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Ok I see

mpqp/core/circuit.py
hJaffaliColibritd: Don't count on Braket Error ,and raise ourselves error if the measurement is not at the end

JulienCalistoTD: Done: [bafbbd1](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/bafbbd128f45b1cfc46798cc1a268987e9cebdc1) add InstructionAfterMeasurementError

mpqp/core/circuit.py
JulienCalistoTD: Done: [bafbbd1](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/bafbbd128f45b1cfc46798cc1a268987e9cebdc1)

mpqp/core/instruction/measurement/pauli_string.py
JulienCalistoTD: Done: [bafbbd1](https://github.com/ColibrITD-SAS/mpqp/pull/165/commits/bafbbd128f45b1cfc46798cc1a268987e9cebdc1) renamed `QISKIT` to `QISKIT_COLORING_GREEDY`

hJaffaliColibritd (approved):

PR #166, opened 2025-08-26
feat: direct imports

Henri-ColibrITD:


docs/barrier.rst
```diff
@@ -1,9 +1,4 @@
 Barriers
 ========
 
-.. code-block:: python
-    :class: import
-
-    from mpqp.core import Barrier
-
```
hJaffaliColibritd, line 8: Why do we remove imports informations from the documentation?
Because everything would be imported directly from `mpqp` ?

docs/noise.rst
```diff
@@ -16,6 +16,9 @@ presented.
 For more details on how to use noise models taken from real hardware, you can
 look at :ref:`SimulatedDevices`.
 
+Note: you can import all noises from the root ``mpqp`` module, but we also
```
hJaffaliColibritd:
```suggestion
Note: you can import all noise models from the root ``mpqp`` module, but we also
```

docs/barrier.rst
Henri-ColibrITD: Exactly. Should we make that more explicit ? I don't know how

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -285,7 +292,7 @@ def subs(
             new_pauli_string += substituted_monomial
         return new_pauli_string
 
-    def simplify(self, inplace: bool = False) -> PauliString:
+    def simplify(self, inplace: bool = False, round: int = 10) -> PauliString:
```
MoHermes: we could rename the parameter ``round`` to precison or ndigits, so to avoid shadowing the python ``round()`` function

Henri-ColibrITD: done in 32d40de

tests/test_doc.py
```diff
@@ -135,6 +138,12 @@
 from mpqp.tools.pauli_grouping import CommutingTypes, pauli_grouping_greedy
 from mpqp.tools.unitary_decomposition import quantum_shannon_decomposition
 
+theta, k = symbols("θ k")
+obs = Observable(np.array([[0, 1], [1, 0]]))
+circ = QCircuit(
```
JulienCalistoTD: why do we have a QCircuit here ?

Henri-ColibrITD: good question

Henri-ColibrITD: removed in [a2d7691](https://github.com/ColibrITD-SAS/mpqp/pull/166/commits/a2d769175cc8901e6dbd451d0a025fb69362be84)

MoHermes (approved):
Reviewed! good work!

docs/barrier.rst
hJaffaliColibritd: OKok all good. I think if it is clearly appearing in the examples and documentation then no need to precise more.

I would have kept here the imports to show we can important directly from mpqp, but if it is available for most of the main objects, then we can let like this, it will be better

i resolve

JulienCalistoTD (approved):

tests/execution/test_validity.py
```diff
@@ -46,6 +50,13 @@
 s = np.sqrt
 e = np.exp
 
+all_devices = [
+    device
+    for device_family in [IBMDevice, GOOGLEDevice, AWSDevice, ATOSDevice, AZUREDevice]
+    for device in device_family
+]
```
hJaffaliColibritd, line 57: Are we sure we want to run on every device available here ?

tests/core/test_circuit.py
```diff
@@ -259,34 +279,34 @@ def list_myqlm_funky_circuits() -> list[myQLM_Circuit]:
...
+#         assert str(QCircuit(init_param)) == f.read()
+
+
+# @pytest.mark.parametrize(
+#     "init_param",
+#     [1.0, X(0), -1],
+# )
+# def test_init_wrong(init_param: int | Sequence[Instruction]):
+#     with pytest.raises(TypeCheckError):
+#         QCircuit(init_param)
```
hJaffaliColibritd: tests removed ?

mpqp/execution/vqa/__init__.py
```diff
@@ -1,7 +1,4 @@
 # pyright: reportUnusedImport=false
-from sympy import symbols
```
hJaffaliColibritd, line 2: Didnt't we say that we put it here so people can create symbols directly when they import vqa module ? 
Or it is already imported in `mpqp`?

examples/notebooks/1_Basics_of_circuit.ipynb
```diff
@@ -496,10 +497,19 @@
     "scrolled": false
    },
    "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "c:\\Users\\HenrideBoutray\\anaconda3\\Lib\\site-packages\\mpqp\\qasm\\qasm_to_braket.py:134: UnsupportedBraketFeaturesWarning: \n",
```
hJaffaliColibritd: This contains personal info on the user?

mpqp/all.py
```diff
@@ -1,81 +0,0 @@
-# pyright: reportUnusedImport=false
```
hJaffaliColibritd, line 1: Maybe for better transition between versions we could keep mpqp.all to be the same as importing from mpqp? then in future versions slowly removing? or like giving the time to warn users ?

hJaffaliColibritd (approved):

mpqp/execution/vqa/__init__.py
Henri-ColibrITD: it would be nice, but it adds a lot of import time for a small benefit

mpqp/all.py
Henri-ColibrITD: mpqp.all was never advertised officially, I would say it's fine. But we should make a note of the import changes in the changelog

examples/notebooks/1_Basics_of_circuit.ipynb
Henri-ColibrITD:
```suggestion
      "~\\anaconda3\\Lib\\site-packages\\mpqp\\qasm\\qasm_to_braket.py:134: UnsupportedBraketFeaturesWarning: \n",
```

Henri-ColibrITD: good point, fixed in [297c7e1](https://github.com/ColibrITD-SAS/mpqp/pull/166/commits/297c7e1f91b0406f96afb367257a5ddb6602fe79)

tests/core/test_circuit.py
Henri-ColibrITD: I think I removed them because we don't have automated type checking anymore, but I added back manual type checking in this case and I added back the tests (in 836b66e)

mpqp/execution/vqa/__init__.py
hJaffaliColibritd: I agree

PR #171, opened 2025-12-03
Doc auto label

JulienCalistoTD:


.github/labels.yml
```diff
@@ -0,0 +1,55 @@
...
+
+- name: triage
+  description: Need some assessment
+  color: f78a50
+
+- name: top priority
+  description: Must be fixed promptly
+  color: ab8a50
+
+- name: aps
```
Henri-ColibrITD: might not be useful ?

Henri-ColibrITD (changes requested):
noice !

Henri-ColibrITD (approved):

.github/workflows/label.yml
```diff
@@ -0,0 +1,77 @@
+name: Labeler
+on:
+  pull_request:
+    # opened, reopened, and synchronize are default for pull_request
+    # edited - when PR title or body is changed
+    # labeled - when labels are added to PR
+    types: [opened, reopened, synchronize, edited, labeled]
```
MoHermes, line 7: In the Actions page sidebar, it exists two `Labeler`, I think it comes from here, because of `edited` and `labeled` adding extra runs, 

Image

MoHermes (changes requested):

.github/workflows/label.yml
JulienCalistoTD: No, it's just a bug in the PR, it will be removed when this branch is deleted

MoHermes: ok.

MoHermes (approved):

hJaffaliColibritd (approved):

PR #173, opened 2025-12-04
Feat vqa remote

MoHermes:


mpqp/core/circuit.py
```diff
@@ -2109,6 +2129,15 @@ def __repr__(self) -> str:
...
 
+    def transpiled_for_device(self, device: AvailableDevice):
+        if device not in self.transpiled_circuit:
+            self.transpiled_circuit[device] = (
+                self.to_other_device(  # pyright: ignore[reportCallIssue]
+                    device  # pyright: ignore[reportArgumentType]
+                )
+            )
+        return self.transpiled_circuit[device]
+
```
hJaffaliColibritd, line 2180: For some providers, we would call to_other_device for each device , because MPQP translation and transpilation is device specific (like IBM), but for AWS/QLM for instance, we only need to_other_language (which is handled by to_other_device if I remember well), but then would need a different indexing of the dictionnary, otherwise we would export the mpqp circuit many times if we change the AWSDevice for instance (while we could have avoided).

PR #176, opened 2025-12-09
New feature: custom gate symbolic unitary check

Henri-ColibrITD:


mpqp/core/instruction/gates/native_gates.py
```diff
@@ -204,9 +202,7 @@ class RotationGate(NativeGate, ParametrizedGate, SimpleClassReprABC):
 
     def __init__(self, theta: Expr | float, target: int):
         self.parameters = [theta]
-        definition = UnitaryMatrix(
-            self.to_canonical_matrix(), **self.native_gate_options
-        )
+        definition = UnitaryMatrix(self.to_canonical_matrix())
```
hJaffaliColibritd, line 205: Just for curiosity, why did we need `**self.native_gate_options`

Henri-ColibrITD: we use to have `native_gate_options = {"disable_symbol_warn": True}`, which do don't need anymore because we don't have this warning anymore

hJaffaliColibritd (approved):

MoHermes (approved):

PR #182, opened 2026-02-13
Feat: local storage improvements

nm727:
### **Fix 1 : SQL Placeholder Mismatch (save.py)**
**Problem**: The [INSERT INTO jobs] had 6 column names ([type, circuit, device, measure, remote_id, status]) but only 4 ? placeholders. The list-of-jobs branch would crash with [sqlite3.OperationalError] every time. The single-job branch had the inverse issue, only 4 columns, silently dropping remote_id and [status].

**Fix**: Both branches now consistently write all 6 columns with 6 placeholders and pass the full 6-value tuple.

### **Fix 2 : Stale Database Dead-End (setup.py)**
**Problem**: When the DB version didn't match, users got RuntimeError: "please contact library authors" a complete brick wall with no recovery path.

**Fix**: The decorator now automatically:

Backs up the old DB to [<path>.v<old_version>.bak]
Emits a UserWarning explaining what happened
Deletes the stale DB and re-creates a fresh one
Users keep their old data as a backup file and can continue working immediately.

### **Fix 3 : Unsafe [eval()] Deserialization (load.py)**
**Problem**: 20+ bare [eval()] calls on strings from the SQLite database. If a DB file is shared, downloaded, or tampered with, this is arbitrary code execution.

**Fix**: Introduced [_safe_eval()] which:

Builds a restricted namespace containing only known MPQP symbols (from [mpqp.all]) plus numpy helpers, [complex64], [float64])
Sets __builtins__ to {} — blocking access to import, [open], os, exec, __import__, etc.
Wraps every call in try/except with a descriptive [ValueError] on failure
Namespace is lazily built once and cached in [_SAFE_NAMESPACE]
All 20 [eval()] calls are now [_safe_eval()] calls.


nm727:
### Fix 1 : Gate matrix class-level constants
native_gates.py

Promoted [self.matrix = np.array(...)] from per-instance allocation in [__init__] to shared class-level attributes for 8 gates: [Id], [X], [Y], [Z], [H], [S], [S_dagger], [SWAP]. All matrices now use [dtype=np.complex128] explicitly. Verified that [x1.matrix is X.matrix] instances share the class constant (zero per-instance allocation).

### Fix 2 : [Instruction.__eq__] pickle → structural
instruction.py

Replaced [pickle.dumps(self) == pickle.dumps(value)] with:

```python
if not isinstance(value, type(self)):
    return False
return self.to_dict() == value.to_dict()
```

Removed the from pickle import dumps import. This is now consistent with how [Gate.__eq__] and [Breakpoint.__eq__] already worked.

### Fix 3 : Job unit tests
test_job.py

Replaced the 4-line # 3M-TODO placeholder with 276 lines covering 25 test methods across 7 test classes: [TestJobStatus], [TestJobType], [TestJobConstruction], [TestJobStatusProperty], [TestJobEquality], [TestJobRepr], and [TestGenerateJob] (including symbolic substitution).

Henri-ColibrITD:
Hi @nm727, thanks for the work you did!

A few preliminary comments: we don't merge directly do main, because main represents the latest version of MPQP, instead we merge to dev and we merge dev to main only when we are ready to release a new version.

In addition, in order to facilitate the review process, we try to have single feature branches, and merge them one after the other. I see that your work is split into clearly separated fixes, would you be able to submit each of these changes as a separated PR?

Thanks!

nm727:
hello @Henri-ColibrITD thank you for the response! It should be resolved now.

mpqp/local_storage/setup.py
```diff
@@ -49,10 +49,20 @@ def wrapper(*args: Any, **kwargs: dict[str, Any]) -> T:
...
+            backup_path = db_path + f".v{db_version}.bak"
+            shutil.copy2(db_path, backup_path)
+            warn(
+                f"Database version {db_version} is outdated (current: "
+                f"{DATABASE_VERSION}). The old database has been backed up to "
+                f"'{backup_path}' and a fresh database will be created.",
+                stacklevel=2,
+            )
+            Path(db_path).unlink()
+            setup_local_storage(db_path)
```
Henri-ColibrITD, line 65: interesting idea, for now we never changed db version so this problem cannot occur. We might prefer leaving the error as it was and create a proper migration script once we have a new version. To be discussed

JulienCalistoTD:
before merging you need to test the code, if you want to test the local_storage you can do ``` python -m pytest --long .\tests\local_storage\``` and ``` python -m pytest --long .\tests\test_doc.py ```

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1276,11 +1276,11 @@ def to_other_language(
...
             }
             braket_sum = None
-            for target, atom in enumerate(self.atoms):
+            for atom in self.atoms:
                 braket_sum = (
-                    braket_sum @ pauli_gate_map[atom.label](target)
+                    braket_sum @ pauli_gate_map[atom.label]()
                     if braket_sum
-                    else pauli_gate_map[atom.label](target)
+                    else pauli_gate_map[atom.label]()
```
hJaffaliColibritd, line 1283: Why do we want to remove the target ?

nm727: To my understanding, the qubit position is already fully encoded by the tensor-product order. Each atom in self.atoms maps to its qubit position implicitly through the sequential @ composition, with I atoms explicitly kept in place for untouched qubits, so no explicit target index is needed.

PR #183, opened 2026-02-16
Second part

nm727:
### Fix 1 : Gate matrix class-level constants
native_gates.py

Promoted [self.matrix = np.array(...)] from per-instance allocation in [__init__] to shared class-level attributes for 8 gates: [Id], [X], [Y], [Z], [H], [S], [S_dagger], [SWAP]. All matrices now use [dtype=np.complex128] explicitly. Verified that [x1.matrix is X.matrix] instances share the class constant (zero per-instance allocation).

### Fix 2 : [Instruction.__eq__] pickle → structural
instruction.py

Replaced [pickle.dumps(self) == pickle.dumps(value)] with:

```python
if not isinstance(value, type(self)):
    return False
return self.to_dict() == value.to_dict()
```

Removed the from pickle import dumps import. This is now consistent with how [Gate.__eq__] and [Breakpoint.__eq__] already worked.

### Fix 3 : Job unit tests
test_job.py

Replaced the 4-line # 3M-TODO placeholder with 276 lines covering 25 test methods across 7 test classes: [TestJobStatus], [TestJobType], [TestJobConstruction], [TestJobStatusProperty], [TestJobEquality], [TestJobRepr], and [TestGenerateJob] (including symbolic substitution).

### DIFF
--- a/mpqp/core/instruction/gates/native_gates.py
+++ b/mpqp/core/instruction/gates/native_gates.py
@@ -433,11 +433,11 @@ class Id(OneQubitNoParamGate, InvolutionGate):
 
     qlm_aqasm_keyword = "I"
     qiskit_string = "id"
+    matrix = np.eye(2, dtype=np.complex128)
 
     def __init__(self, target: int, label: Optional[str] = None):
         super().__init__(target)
         self.label = label
-        self.matrix = np.eye(2, dtype=np.complex128)
 
     def to_other_language(
         self,
@@ -499,10 +499,10 @@ class X(OneQubitNoParamGate, InvolutionGate):
 
     qlm_aqasm_keyword = "X"
     qiskit_string = "x"
+    matrix = np.array([[0, 1], [1, 0]], dtype=np.complex128)
 
     def __init__(self, target: int):
         super().__init__(target)
-        self.matrix = np.array([[0, 1], [1, 0]])
 
 
 class Y(OneQubitNoParamGate, InvolutionGate):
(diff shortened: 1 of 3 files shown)


mpqp/core/instruction/gates/native_gates.py
```diff
@@ -433,11 +433,11 @@ def cirq_gate(cls):
 
     qlm_aqasm_keyword = "I"
     qiskit_string = "id"
+    matrix = np.eye(2, dtype=np.complex128)
```
Henri-ColibrITD, line 436: I think we don't do this because some of the gates' matrix can only be instantiated at `__init__` time, so we try to stay homogeneous

Henri-ColibrITD:
In this PR, you still have 3 relatively unrelated changes, could you split them into 3 PRs ?
 
It may sound a bit strange to have a 4 lines changed PR for your Fix 2, but for instance this one would be the one change we can accept without changing anything, while the two other modifications we will likely have to retouch them

nm727:
Closing this in favor of the three separate PRs requested: PR #184 PR #185 & PR #186

mpqp/core/instruction/gates/native_gates.py
nm727: I see your point regarding consistency. I took a closer look at the current inheritance structure to see how we could best align these changes with the rest of the library.

From what I can see, the gates whose matrices depend on parameters (like Rx, Rz, or P) inherit from RotationGate. These don’t actually use self.matrix, but instead override to_canonical_matrix() to calculate the matrix from self.parameters.

Within NoParameterGate, the pattern seems to vary. For example, T, CNOT, and CZ also override the method directly and don't set a self.matrix attribute. It seemed that only the 8 constant gates I modified were using self.matrix in __init__. Since their matrices are true constants, I thought moving them might simplify the initialization.

However, I might have missed a specific convention you want to maintain here. If you feel that keeping self.matrix inside __init__ is better for readability or consistency with future plans, I’m more than happy to revert that part!

PR #185, opened 2026-02-16
Update instruction logic

nm727:


Henri-ColibrITD (approved):

hJaffaliColibritd (approved):

mpqp/core/instruction/instruction.py
```diff
@@ -95,7 +94,9 @@ def to_other_language(
         pass
 
     def __eq__(self, value: object) -> bool:
-        return dumps(self) == dumps(value)
+        if not isinstance(value, type(self)):
+            return False
+        return self.to_dict() == value.to_dict()
```
MoHermes: Using ``to_dict()`` makes equality more stable compared to ``pickle.dumps()`` .  Just a minor tweak is to return ``NotImplemented`` rather than ``False``, also consider naming ``other``.

Henri-ColibrITD: agreed for `other`, not so much for `NotImplemented` ^^

MoHermes (approved):

PR #186, opened 2026-02-16
Tests: jobs status and inner logic

nm727:


tests/execution/test_job.py
```diff
@@ -1,4 +1,271 @@
...
+    def test_all_statuses_exist(self):
+        expected = {"INIT", "QUEUED", "RUNNING", "CANCELLED", "ERROR", "DONE"}
+        actual = {s.name for s in JobStatus}
+        assert expected == actual
+
+    def test_terminal_statuses(self):
+        """DONE, ERROR, CANCELLED should be considered terminal."""
+        terminal = {JobStatus.DONE, JobStatus.ERROR, JobStatus.CANCELLED}
+        non_terminal = {JobStatus.INIT, JobStatus.QUEUED, JobStatus.RUNNING}
+        assert terminal.isdisjoint(non_terminal)
```
Henri-ColibrITD: unsure about the usefulness of these tests

tests/execution/test_job.py
```diff
@@ -1,4 +1,271 @@
...
+        job = Job(JobType.STATE_VECTOR, circuit, IBMDevice.AER_SIMULATOR)
+        assert job._status == JobStatus.INIT
+
+    def test_construction_with_measure(self):
+        circuit = QCircuit([H(0), CNOT(0, 1)])
+        measure = BasisMeasure([0, 1], shots=100)
+        job = Job(JobType.SAMPLE, circuit, IBMDevice.AER_SIMULATOR, measure)
+        assert job.measure is not None
+        # measure should be deep-copied
+        assert job.measure is not measure
```
Henri-ColibrITD: I like it

tests/execution/test_job.py
```diff
@@ -1,4 +1,271 @@
...
+        job.status = JobStatus.RUNNING
+        assert job._status == JobStatus.RUNNING
+
+    def test_local_status_returns_directly(self):
+        """For local (non-remote) devices, the status property should return
+        the stored status without attempting a remote call."""
+        circuit = QCircuit(2)
+        job = Job(JobType.STATE_VECTOR, circuit, IBMDevice.AER_SIMULATOR)
+        job.status = JobStatus.DONE
+        # Should not raise — no remote lookup attempted for local device
```
Henri-ColibrITD: the mentioned feature is not tested IMO

tests/execution/test_job.py
```diff
@@ -1,4 +1,271 @@
...
+# Job status property
+# ---------------------------------------------------------------------------
+
+
+class TestJobStatusProperty:
+    def test_status_setter(self):
+        circuit = QCircuit(2)
+        job = Job(JobType.STATE_VECTOR, circuit, IBMDevice.AER_SIMULATOR)
+        job.status = JobStatus.RUNNING
+        assert job._status == JobStatus.RUNNING
```
Henri-ColibrITD: not a super fan of testing setters and init unless they have complex behavior. It adds testing time, makes the overall architecture harder to change, and doesn't really add useful securities to the testing suite

tests/execution/test_job.py
```diff
@@ -1,4 +1,271 @@
...
+        assert job.status == JobStatus.DONE
+
+    def test_terminal_status_no_remote_check(self):
+        """Once a job reaches a terminal state, the status property should
+        return immediately for any device."""
+        circuit = QCircuit(2)
+        job = Job(JobType.STATE_VECTOR, circuit, IBMDevice.AER_SIMULATOR)
+        for terminal in [JobStatus.DONE, JobStatus.ERROR, JobStatus.CANCELLED]:
+            job.status = terminal
+            assert job.status == terminal
```
Henri-ColibrITD: same ?

tests/execution/test_job.py
```diff
@@ -1,4 +1,271 @@
...
+        job1 = Job(JobType.STATE_VECTOR, circ1, IBMDevice.AER_SIMULATOR)
+        job2 = Job(JobType.STATE_VECTOR, circ2, IBMDevice.AER_SIMULATOR)
+        assert job1 != job2
+
+    def test_not_equal_to_non_job(self):
+        circuit = QCircuit(2)
+        job = Job(JobType.STATE_VECTOR, circuit, IBMDevice.AER_SIMULATOR)
+        assert job != "not a job"
+        assert job != 42
+        assert job != None
```
Henri-ColibrITD: maybe we can use a `pytest.mark.parametrize` for these tests ?

tests/execution/test_job.py
```diff
@@ -1,4 +1,271 @@
...
+        circuit = QCircuit(2)
+        job = Job(JobType.STATE_VECTOR, circuit, IBMDevice.AER_SIMULATOR)
+        d = job.to_dict()
+        assert "job_type" in d
+        assert "circuit" in d
+        assert "device" in d
+        assert "measure" in d
+        assert "id" in d
+        assert "status" in d
+
```
Henri-ColibrITD: these tests look weird to me, why not testing the whole repr at this point ?

tests/execution/test_job.py
```diff
@@ -1,4 +1,271 @@
...
+    def test_symbolic_substitution(self):
+        from sympy import symbols
+
+        theta = symbols("θ")
+        from mpqp.gates import Rx
+
+        circuit = QCircuit([Rx(theta, 0)])
+        job = generate_job(circuit, IBMDevice.AER_SIMULATOR, {theta: 1.5})
+        # After substitution the circuit should have no free variables
+        assert len(job.circuit.variables()) == 0
```
Henri-ColibrITD, line 253: nice

tests/execution/test_job.py
Henri-ColibrITD, line 1: In general, I like what you did, but I think the important thing to test in this file is mostly the remote thing, we can merge this once you reply to/take into account the comments, but the biggest issue remains :(

Also, a general question: what are the classes for ? I never used them so I'm curious :)

nm727:
@Henri-ColibrITD Thank you for the thorough review! I addressed most of your concerns in this commit and attempted at the remote tests.

As for the test classes, they are mainly for grouping, pytest uses them to namespace the -v output, so instead of seeing:
```
test_local_device_never_polls PASSED
test_no_measurement_gives_state_vector PASSED
```
you see:

```
TestRemoteStatus::test_local_device_never_polls PASSED
TestGenerateJob::test_no_measurement_gives_state_vector PASSED
```

It makes it easier to see which area failed when scanning test output. Totally optional, just a small structural choice to keep things organized as the suite grows

tests/execution/test_job.py
MoHermes: here we can remove the tests, no need.

tests/execution/test_job.py
```diff
@@ -1,4 +1,269 @@
...
+        circuit = QCircuit(2)
+        devices = [
+            IBMDevice.AER_SIMULATOR,
+            ATOSDevice.MYQLM_PYLINALG,
+            AWSDevice.BRAKET_LOCAL_SIMULATOR,
+            GOOGLEDevice.CIRQ_LOCAL_SIMULATOR,
+        ]
+        for device in devices:
+            job = Job(JobType.STATE_VECTOR, circuit, device)
+            assert job.device == device
```
MoHermes: I would remove this test, just looping through devices to assert job.device is device

tests/execution/test_job.py
```diff
@@ -1,4 +1,269 @@
...
+# ---------------------------------------------------------------------------
+# JobType enum
+# ---------------------------------------------------------------------------
+
+
+class TestJobType:
+    def test_all_types_exist(self):
+        expected = {"STATE_VECTOR", "SAMPLE", "OBSERVABLE"}
+        actual = {t.name for t in JobType}
+        assert expected == actual
```
MoHermes: remove this one, doesn't test behaviour

tests/execution/test_job.py
MoHermes: Agree here we avoid asserting private/internal fields in tests

nm727:
should be resolved

tests/execution/test_job.py
Henri-ColibrITD: the issue is not necessarily about private fields (even if indeed it's an issue), but more generally about tests in the shape of
```py
a = 1
assert a == 1
```
Those kinds of tests bring no value to a project

hJaffaliColibritd:
Dear @nm727 ,

We took time to have internal discussion about the proposed test,

At the end we will not integrate them with the current code base, but we would be happy to see other contributions from your side in the future. We will think internally on how to conduct properly testing on remote jobs.

Thank you again

PR #189, opened 2026-03-11
Fix pauli grouping doc

MathieuG-Colibri:
Fixed docs for plot and pauli grouping as well as removing warnings raised during tests.


docs/conf.py
```diff
@@ -51,6 +51,19 @@
...
+    "BatchResult": "BatchResult",
+    "NoiseModel": "NoiseModel",
+    "AnswerNode": "AnswerNode",
+    "Qubo": "Qubo",
+    "Instruction": "Instruction",
+    "Sample": "Sample",
+    "Instr": "Instr",
+    "UserGate": "UserGate",
+    "SingleQubitGate": "SingleQubitGate",
+    "CustomGate": "CustomGate",
```
Henri-ColibrITD, line 67: could we at some point build this list automagically ? maybe not in this PR but if you like the idea could be an upcoming task (not needed for next release)

MathieuG-Colibri: I'm putting the task on the backlog, I'll take a look at it later.

docs/execution.rst
```diff
@@ -45,4 +45,20 @@ Jobs
 Results
 -------
 
-.. automodule:: mpqp.execution.result
\ No newline at end of file
+.. automodule:: mpqp.execution.result
```
Henri-ColibrITD: it would be nice to have the pictures directly in the docstring, but I guess if you chose this option it's because there is no way to include pictures in docstrings ?

mpqp/core/instruction/gates/gate_definition.py
```diff
@@ -224,7 +224,6 @@ def caster(v: Expr | Complex) -> Expr | Complex:
             or not any(isinstance(val, Expr) for _, val in np.ndenumerate(matrix))
             else object
         )
-
```
Henri-ColibrITD, line 231: ? x)

mpqp/core/instruction/gates/parametrized_gate.py
```diff
@@ -63,9 +63,19 @@ def subs(
         concrete_gate.definition = concrete_gate.definition.subs(
             values, remove_symbolic, **options
         )
-        caster = lambda v: float(v) if remove_symbolic else v
+
+        def caster(v: Expr | float) -> float:
```
Henri-ColibrITD: caster is not supposed to remove the symbols if `remove_symbolic` is `False`

mpqp/execution/providers/aws.py
```diff
@@ -517,7 +519,7 @@ def get_result_from_aws_task_arn(task_arn: str) -> Result:
         AWSBraketRemoteExecutionError: When the status of the task is unknown.
     """
     from braket.aws import AwsQuantumTask
-    from braket.tasks import GateModelQuantumTaskResult, QuantumTask
+    from braket.tasks import GateModelQuantumTaskResult
 
     task: QuantumTask = AwsQuantumTask(task_arn)
```
Henri-ColibrITD, line 545: having this annotation doesn't pose problem without the import ?

mpqp/execution/providers/ibm.py
Henri-ColibrITD, line 1: this is a bit complicated, can you spend some time explaining the changes say tuesday ?

mpqp/execution/result.py
Henri-ColibrITD, line 1: love it

Henri-ColibrITD (changes requested):
a few comments but overall very good job

mpqp/execution/providers/aws.py
MathieuG-Colibri: Nope because of the added from __future__ import annotations.

It made the type checker angry

mpqp/execution/providers/ibm.py
MathieuG-Colibri: This is really easy it's just that github freaked out.

So there were a warning when applying the noise in qiskit, everything was caught using a try catch and a logger.

Since I removed the warning when applying the noise hence removed the try catch.

docs/execution.rst
MathieuG-Colibri: It didn't seem possible so I went with this ugly fix, but it works nicely.

mpqp/core/instruction/gates/parametrized_gate.py
MathieuG-Colibri: It is true however an issue rised when we try to sub without remove_symbolic as the type don't return to a clean python type but it stays as an ugly sympy expression. Hence it breaks later in the execution when we try running the circuit.

docs/execution.rst
Henri-ColibrITD: fixed in 6eee200

mpqp/core/instruction/gates/parametrized_gate.py
Henri-ColibrITD: this potentially changes behavior, but from what I gathered, the issue might from from the fact that, just before execution, we should run `subs({}, remove_symbolic=True)` to avoid having sympy numeric expressions left over. (I'm not a fan of the solution consisting in systematically replacing numeric expressions by floats, because you might replace sympy's `pi` by 3.1415..., which would 1. lose precision (minor issue) 2. display ugly (other minor issue), both of these issues are minor but avoidable)

Henri-ColibrITD: after some thoughts, none of those solutions are perfect, but I don't think anymore than the old one is necessary better. The only thing is, if you decide to stick to the new one, I think you can completely delete `remove_symbolic` from the function signature

MoHermes (approved):

PR #197, opened 2026-05-06
Perf multiple circuit run for one job

JulienCalistoTD:


JulienCalistoTD:
transform jobs to single job but with multiple circuits

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -198,6 +198,14 @@ def diagonal_elements(self, diag_elements: list[float] | npt.NDArray[np.float64]
         self._pauli_string = None
         self._matrix = None
 
+    def is_matrix(self):
+        """Returns if the matrix is already set in the observable."""
+        return not self._matrix is None
+
+    def is_pauli(self):
```
hJaffaliColibritd, line 205: I would rename the function `is_pauli` or `is_matrix` which can be misleading. Maybe `is_matrix_set` / `is_pauli_string_set` can do the job?

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd, line 40: Maybe move this to `mpqp.tools.generics` ?
Or at least not put it in TYPE_CHECKING so it can be used in isinstance

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1124,6 +1124,11 @@ def __imul__(self, other: "Coef") -> PauliStringMonomial:
         return self
 
     def __itruediv__(self, other: "Coef") -> PauliStringMonomial:
+        from sympy import Expr
+
+        assert isinstance(
+            other, (int, float, complex, Expr)
```
hJaffaliColibritd, line 1130: We can't directly use `Coef` here? I remember we already had this discussion/issue for `Matrix`, but in principle if `Coef` is also a Union of types, it should be able to check it

mpqp/core/circuit.py
```diff
@@ -1813,6 +1834,30 @@ def from_other_language(
...
+
+        for measurement in self.measurements:
+            if isinstance(measurement, BasisMeasure):
+                if measurement.shots <= 0:
+                    return JobType.STATE_VECTOR
+                else:
+                    return JobType.SAMPLE
+            elif isinstance(measurement, ExpectationMeasure):
+                return JobType.OBSERVABLE
+        return JobType.STATE_VECTOR
```
hJaffaliColibritd, line 1852: Note for me: double check that this is used in `generate_jobs` when we call `run`, to no do the same job twice

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,382 @@ def variables(self) -> set[Basic]:
...
         return params
+
+
+class BindingMode(Enum):
+    PRODUCT = auto()
+    ZIP = auto()
+
+
+class CircuitBinding:
+    def __init__(
```
hJaffaliColibritd, line 1996: I would have put it in another file than circuit.py, if it doesn't imply circular dependencies

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,382 @@ def variables(self) -> set[Basic]:
...
+        self,
+        circuits: OneOrMany[QCircuit | CircuitBinding],
+        values: Optional[OneOrMany[dict[Expr | str, Complex | float]]] = None,
+        measurements: Optional[OneOrMany[Measure]] = None,
+        mode: BindingMode = BindingMode.PRODUCT,
+        noises: Optional[list[NoiseModel]] = None,
```
hJaffaliColibritd, line 2002: This also be combined with product/zip (different noise models for different circuits/values/measurements?) ? If not then maybe just use the one attached to the `QCircuit`, I don't see why would we need to put it as a separate parameter

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,382 @@ def variables(self) -> set[Basic]:
...
+        shots: Optional[int] = None,
```
hJaffaliColibritd, line 2003: Same remark for the shots, in principle they should be already informed in the Measure.

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,382 @@ def variables(self) -> set[Basic]:
...
                         params.update(param.free_symbols)
         return params
+
+
+class BindingMode(Enum):
+    PRODUCT = auto()
+    ZIP = auto()
+
+
+class CircuitBinding:
```
hJaffaliColibritd, line 1995: Documentation, with examples

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,382 @@ def variables(self) -> set[Basic]:
                     if isinstance(param, Expr):
                         params.update(param.free_symbols)
         return params
+
+
+class BindingMode(Enum):
```
hJaffaliColibritd, line 1990: Documentation

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,382 @@ def variables(self) -> set[Basic]:
...
+    def unroll(
+        self,
+    ) -> list[
+        tuple[
+            QCircuit,
+            Optional[dict["Expr | str", "Complex | float"]],
+            Optional["Measure"],
+        ]
+    ]:
+        """Resolves the lazy execution graph and returns a flat list of
```
hJaffaliColibritd, line 2198: Better documentation

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,382 @@ def variables(self) -> set[Basic]:
...
+            else ([self.measurements] if self.measurements is not None else [None])
+        )
+
+        def merge_vals(v_base, v_curr):
+            if v_base is None and v_curr is None:
+                return None
+            merged = dict(v_base) if v_base is not None else {}
+            if v_curr is not None:
+                merged.update(v_curr)
+            return merged
```
hJaffaliColibritd, line 2228: could be moved to tools? why defined inside the method ?

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,382 @@ def variables(self) -> set[Basic]:
...
+    def to_other_device(
+        self, device: IBMDevice
+    ) -> list[tuple["EstimatorPubLike", "Job"]]: ...
+    @overload
+    def to_other_device(
+        self, device: AvailableDevice, programSet: Literal[False]
+    ) -> "CircuitBinding": ...
+    def to_other_device(
+        self, device: AvailableDevice, programSet: bool = True
+    ) -> "CircuitBinding | tuple[ProgramSet, list[tuple[Any]]] | list[tuple[EstimatorPubLike, Job]]":
```
hJaffaliColibritd, line 2286: Maybe use same organization as we recently did. Move this to mpqp.translation and call it from here ?

mpqp/execution/providers/aws.py
```diff
@@ -143,7 +144,130 @@ def run_braket(job: Job) -> Result:
...
+                    result = task[index][0]
+                    for name, eigenvalue in eigenvalues.items():
+                        for i in range(length):
+                            binary_state = f"{bin(i)[2:].zfill(len(bin(length))- 3)}"
+                            if binary_state in result.probabilities:
+                                sorted_values[i] = result.probabilities[
+                                    binary_state
+                                ].real
+                            else:
+                                sorted_values[i] = 0
```
hJaffaliColibritd, line 212: All the grouping thing I would create another method to handle that, for circuit biding or not circuit biding, and even I would say braket or not braket, because it is a general process that can be reused, right ? 
Otherwise we will have as much implementations of the pauli grouping as providers and as ways to run circuits with observables (run normal, run binded, etc)

To be discussed

mpqp/execution/job.py
```diff
@@ -129,9 +133,17 @@ def __init__(
         """Optional message associated with the current job status, especially
         for execution errors."""
 
+        """Store the measurement of a circuit."""
```
hJaffaliColibritd, line 136: I think this is more used for circuit biding right ?

mpqp/execution/job.py
```diff
@@ -129,9 +133,17 @@ def __init__(
...
+        self.measurement = measurement
+
+        """Store parameters in case the original circuit was parametrized."""
```
hJaffaliColibritd, line 139: Or maybe say defined by symbolic parameters

mpqp/execution/job.py
```diff
@@ -129,9 +133,17 @@ def __init__(
...
+        self.values = values
+
     @property
     def measure(self) -> Optional[Measure]:
         """Returns the first measurement from the circuit's measurements."""
+        if isinstance(self.circuit, CircuitBinding):
+            return None
```
hJaffaliColibritd, line 146: here you would return self.measurement in the case of a CircutBiding no ?

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,472 @@ def variables(self) -> set[Basic]:
...
+            pubs_by_circuit = {}
+            for c, v, m in unrolled:
+                c_id = id(c)
+                if c_id not in pubs_by_circuit:
+                    pubs_by_circuit[c_id] = {"circuit": c, "params": [], "measures": []}
+
+                params = [str(var) for var in c.variables()]
+                if v:
+                    values = [v[key] for key in params if key in v.keys()]
+                else:
```
MoHermes: Two issues here:
binding order from ``c.variables()`` can differ from qiskit’s parameter order, so values may be bound to the wrong parameters and produce incorrect expectation values
Also,
``str(var)`` breaks bindings that use SymPy symbols as dictionary keys

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,472 @@ def variables(self) -> set[Basic]:
...
+                        pubs_with_context.append((pub, context_job))
+                    continue
+
+                if all(len(p) == 0 for p in params):
+                    params = None
+                if all(len(o) == 0 for o in q_obs):
+                    q_obs = None
+                if q_obs and params:
+
+                    pub = (q_c, q_obs, params)
```
MoHermes, line 2435: I tested with two parameter sets and Z and X observables, and it produces incorrect expectation values
I think it comes from how q_obs and params are structured when building the PUB, that makes qiskit broadcast the parameter sets with the observables incorrectly

mpqp/execution/providers/ibm.py
```diff
@@ -622,47 +882,108 @@ def extract_result(
...
 
-            # If only one result, we directly return the expectation value
-            if len(exp_values) == 1:
-                return Result(job, float(exp_values[0]), float(stds[0]), shots)
+            if exp_values.ndim == 0:
+                val = float(exp_values)
+                std = float(stds) if stds.size > 0 else 0.0
+                return Result(job, val, std, shots)
+            elif exp_values.ndim == 2:
+                N_obs, M_params = exp_values.shape
```
MoHermes, line 924: The result here is assumed to be observables on (axis 0) parameter sets on (axis 1), but that depends on how qiskit broadcasts the PUB inputs and with the current PUB construction, this can map values to the wrong results

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,472 @@ def variables(self) -> set[Basic]:
...
+
+        self.job_type = circuits[0].job_type
+        self.nb_qubits = circuits[0].nb_qubits
+        for circ in circuits:
+            if circ.job_type != self.job_type:
+                if (
+                    circ.job_type == JobType.STATE_VECTOR
+                    or self.job_type == JobType.STATE_VECTOR
+                ):
+                    self.job_type = circ.job_type
```
MoHermes, line 2074: The current behavior chooses one job_type for the whole binding based on circuit order, the same pair of circuits is handled differently depending only on their order
for example:
If I have two circuits, the first one represents state-vector and the second represents sample, I got:
If the order is (state-vector, sample), binding becomes SAMPLE, then AssertionError
And if I reverse the order,  binding becomes STATE_VECTOR, then the sampling circuit returns a state vector instead of samples

mpqp/execution/runner.py
```diff
@@ -349,9 +363,64 @@ def _run_single(
         raise NotImplementedError(f"Device {device} not handled")
 
 
+def _run_circuit_binding(
+    circuit_binding: CircuitBinding,
+    device: AvailableDevice,
+    display_breakpoints: bool = True,
```
MoHermes: missing ``provider_params``

mpqp/execution/runner.py
```diff
@@ -465,24 +534,53 @@ def namer(circ: QCircuit, i: int):
...
+                            namer(circ, i + 1),
+                            dev,
+                            values,
+                            display_breakpoints,
+                        )
+                    )
+            else:
+                if values is not None:
+                    raise ValueError("values must be specified in CircuitBinding")
+                return _run_circuit_binding(circuit, dev, display_breakpoints)
```
MoHermes: This returns after the first device execution, so for multiple devices ``CircuitBinding `` stops after the first device

mpqp/core/circuit.py
```diff
@@ -1940,3 +1985,472 @@ def variables(self) -> set[Basic]:
...
+                            measure.shots = self.shots
+                else:
+                    if shots_ is None:
+                        shots_ = measure.shots
+                    elif measure.shots != shots_:
+                        raise ValueError(
+                            "All measurements in CircuitBinding must have the same number of shots."
+                        )
+                if isinstance(measure, ExpectationMeasure):
+                    measurements[index] = adjust_measure(measure, self.nb_qubits)
```
MoHermes, line 2143: ``self.nb_qubits = max(self.nb_qubits, circ.nb_qubits)``, so:
For example, when both circuits use the same ``Z `` measurement, the expected measurement will be ``Z`` for c1(1-qubit) and ``Z x I`` for c2 (2-qubits)
So, reuse the new padded measurement ``Z x I`` for both circuits, makes 1-qubit circuit gets a 2-qubit observable and fails with:
``ValueError: The number of qubits of the circuit (1) does not match the number of qubits of the (0, 0)-th observable (2).``

I think the measurement should be adjusted separately for each circuit width, not once to the max width

mpqp/execution/runner.py
MoHermes, line 350: not missing ``provider_parmas``?

mpqp/core/circuit.py
MathieuG-Colibri: Normally we have a limit that every circuits in a CB should be of the same nbr of qubits. So having adjust measure shouldn't be an issue.

mpqp/core/circuit.py
MathieuG-Colibri: That's alright because we cannot have different job types inside of a circuitbinding.
So either order will juste result in an error raised.

mpqp/core/circuit.py
MathieuG-Colibri: First issue is going to be fixed.

However second is not an issue because it's only us (mpqp) that uses sympy symbols as key most other providers only supports str keys.

mpqp/core/circuitbinding.py
```diff
@@ -0,0 +1,709 @@
...
+                        "your circuit already contains measurements, "
+                        "you cannot have multiple measurements"
+                    )
+
+                circuit_measurements = circuit.measurements
+                circuit_without_measurements = circuit.without_measurements(
+                    deep_copy=False
+                )
+
+                circuit_without_measurements.transpiled_circuit = None
```
hJaffaliColibritd: Why can't we reuse an already transpiled circuit ?

PR #198, opened 2026-05-17
Feat adjust measure

MathieuG-Colibri:
Updated the way to treat expectation measures.

If the measure doesn't cover the whole registry we pad the pauli string with Identities, or is not ordered (instead of swap gates reorganize the matrix or pauli string).

Isn't necessary for braket with optimize_measure set at true (does the pauli grouping).

### DIFF
--- a/mpqp/core/instruction/measurement/expectation_value.py
+++ b/mpqp/core/instruction/measurement/expectation_value.py
@@ -8,10 +8,10 @@ from __future__ import annotations
 import copy
 from numbers import Real
 from typing import TYPE_CHECKING, Literal, Optional, Union, overload
-from typing_extensions import Never
 
 import numpy as np
 import numpy.typing as npt
+from typing_extensions import Never
 
 from mpqp.core.instruction.measurement.measure import Measure
 from mpqp.core.instruction.measurement.pauli_string import (
@@ -34,8 +34,6 @@ if TYPE_CHECKING:
     from qiskit._accelerate.circuit import Parameter
     from qiskit.quantum_info import SparsePauliOp
     from sympy import Expr
-
-    from mpqp.core.instruction.gates.custom_controlled_gate import Gate
 
 
 class Observable:
@@ -334,7 +332,7 @@ class Observable:
             return QLMObservable(self.nb_qubits, matrix=self.matrix)
         elif language == Language.BRAKET:
             if self._pauli_string:
-                from braket.circuits.observables import TensorProduct, Sum
+                from braket.circuits.observables import Sum, TensorProduct
 
                 obs = self.pauli_string.to_other_language(Language.BRAKET)
                 if isinstance(obs, TensorProduct):
@@ -466,9 +464,7 @@ class ExpectationMeasure(Measure):
             self.observables.append(new_obs)
 
         if targets is None:
-            self.targets = list(range(observable[0].nb_qubits))
-        self.rearranged_targets = list(self.targets)
-        self._pre_measure: list[Gate] = []
+            self.targets = list(range(self.observables[0].nb_qubits))
 
     @property
     def nb_observables(self) -> int:
@@ -477,10 +473,6 @@ class ExpectationMeasure(Measure):
     @property
     def observables_labels(self) -> list[str]:
         return [o.label for o in self.observables if o.label is not None]
-
-    @property
-    def pre_measure(self) -> list[Gate]:
-        return self._pre_measure
 
     def get_pauli_grouping(self) -> list[list[PauliStringMonomial]]:
         """Return the grouped monomials of the Pauli string of the observable.
--- a/mpqp/core/instruction/measurement/expectation_value.py
+++ b/mpqp/core/instruction/measurement/expectation_value.py
@@ -9,12 +9,10 @@ import copy
 from numbers import Real
 from typing import TYPE_CHECKING, Literal, Optional, Union, overload
 from typing_extensions import Never
-from warnings import warn
 
 import numpy as np
 import numpy.typing as npt
 
-from mpqp.core.instruction.gates.native_gates import SWAP
 from mpqp.core.instruction.measurement.measure import Measure
 from mpqp.core.instruction.measurement.pauli_string import (
     CommutingTypes,
@@ -484,47 +482,8 @@ class ExpectationMeasure(Measure):
         """Ensures target qubits are ordered and contiguous, rearranging them if
         necessary (private)."""
 
-        if len(self.targets) == 0:
-            self._pre_measure: list[Gate] = []
-            return
-
-        if self.nb_qubits != self.observables[0].nb_qubits:
-            raise NumberQubitsError(
-                f"Target size {self.nb_qubits} doesn't match observable size "
-                f"{self.observables[0].nb_qubits}."
-            )
-
+        self.rearranged_targets = list(self.targets)
         self._pre_measure: list[Gate] = []
-        """List of Gates added before the expectation measurement to correctly swap
-        target qubits when their are not ordered or contiguous."""
-        targets_is_ordered = all(
-            [self.targets[i] > self.targets[i - 1] for i in range(1, len(self.targets))]
-        )
-        tweaked_tgt = copy.copy(self.targets)
-        if (
-            max(tweaked_tgt) - min(tweaked_tgt) + 1 != len(tweaked_tgt)
-            or not targets_is_ordered
-        ):
-            warn(
-                "Non contiguous or non sorted observable target will introduce "
-                "additional CNOTs."
-            )
-
-            for t_index, target in enumerate(tweaked_tgt):  # sort the targets
-                min_index = tweaked_tgt.index(min(tweaked_tgt[t_index:]))
-                if t_index != min_index:
-                    self._pre_measure.append(SWAP(target, tweaked_tgt[min_index]))
-                    tweaked_tgt[t_index] = tweaked_tgt[min_index]
-                    tweaked_tgt[min_index] = target
-            for t_index, target in enumerate(tweaked_tgt):  # compact the targets
-                if t_index == 0:
-                    continue
-                if target != tweaked_tgt[t_index - 1] + 1:
-                    self._pre_measure.append(SWAP(target, tweaked_tgt[t_index - 1] + 1))
-                    tweaked_tgt[t_index] = tweaked_tgt[t_index - 1] + 1
-        self.rearranged_targets = tweaked_tgt
-        """Adjusted list of target qubits when they are not initially sorted and
-        contiguous."""
 
     @property
     def pre_measure(self) -> list[Gate]:
--- a/mpqp/core/instruction/measurement/expectation_value.py
+++ b/mpqp/core/instruction/measurement/expectation_value.py
@@ -467,7 +467,8 @@ class ExpectationMeasure(Measure):
 
         if targets is None:
             self.targets = list(range(observable[0].nb_qubits))
-        self._check_targets_order()
+        self.rearranged_targets = list(self.targets)
+        self._pre_measure: list[Gate] = []
 
     @property
     def nb_observables(self) -> int:
@@ -476,13 +477,6 @@ class ExpectationMeasure(Measure):
     @property
     def observables_labels(self) -> list[str]:
         return [o.label for o in self.observables if o.label is not None]
-
-    def _check_targets_order(self):
-        """Ensures target qubits are ordered and contiguous, rearranging them if
-        necessary (private)."""
-
-        self.rearranged_targets = list(self.targets)
-        self._pre_measure: list[Gate] = []
 
     @property
     def pre_measure(self) -> list[Gate]:
--- a/mpqp/core/instruction/measurement/expectation_value.py
+++ b/mpqp/core/instruction/measurement/expectation_value.py
@@ -22,7 +22,6 @@ from mpqp.core.instruction.measurement.pauli_string import (
 )
 from mpqp.core.languages import Language
 from mpqp.tools.display import one_lined_repr
-from mpqp.tools.errors import NumberQubitsError
 from mpqp.tools.generics import Matrix
 from mpqp.tools.maths import is_diagonal, is_hermitian, is_power_of_two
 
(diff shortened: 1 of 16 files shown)


mpqp/execution/providers/ibm.py
```diff
@@ -106,6 +106,9 @@ def compute_expectation_value(
         if TYPE_CHECKING:
             assert isinstance(translated, SparsePauliOp)
         qiskit_observables.append(translated)
+        qiskit_observables = [
+            obs.apply_layout(ibm_circuit.layout) for obs in qiskit_observables
+        ]
```
Henri-ColibrITD: can we do this in a single step ?

mpqp/execution/runner.py
```diff
@@ -55,14 +55,16 @@
...
     """We allow the measure to not span the entire circuit, but providers
     usually do not support this behavior. To make this work, we tweak the measure
     this function to match the expected behavior.
 
     In order to do this, we add identity measures on the qubits not targeted by
-    the measure. In addition to this, some swaps are automatically added so the
-    the qubits measured are ordered and contiguous (though this is done in
-    :func:`generate_job`)
+    the measure. pauli observables are directly embeded on their target qubits,
+    while matrix observables are padded with identity matrices when the targets
```
Henri-ColibrITD: is the id padding in the matrix case a regression ?

mpqp/tools/circuit.py
```diff
@@ -86,6 +87,12 @@ def random_circuit(
     qcircuit = QCircuit(nb_qubits)
     for _ in range(nb_gates):
         qcircuit.add(random_gate(gate_classes, nb_qubits, rng))
+
+    if use_all_qubits:  # used in case we want to test braket
+        from mpqp.gates import H
+
+        for i in range(nb_qubits):
+            qcircuit.add(H(i))
```
Henri-ColibrITD: maybe `random_gate` instead of H ?

mpqp/tools/maths.py
```diff
@@ -478,3 +479,135 @@ def is_power_of_two(n: int) -> bool:
...
+    >>> pprint(matrix)
+    [[0, 1, 0, 0],
+     [1, 0, 0, 0],
+     [0, 0, 0, 1],
+     [0, 0, 1, 0]]
+    >>> pprint(rearrange_matrix(matrix, [1,0]))
+    [[0, 0, 1, 0],
+     [0, 0, 0, 1],
+     [1, 0, 0, 0],
+     [0, 1, 0, 0]]
```
Henri-ColibrITD: starting with a diagonal matrix might be more pedagogical/easier to understand what happens

mpqp/tools/circuit.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/198/commits/e8b7d912640134196d0c0d18286dd6f9667b8760

mpqp/tools/maths.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/198/commits/e8b7d912640134196d0c0d18286dd6f9667b8760

mpqp/execution/runner.py
MathieuG-Colibri: Is it still needing discussion ?

mpqp/execution/providers/ibm.py
MathieuG-Colibri: It was deadcode, now deleted https://github.com/ColibrITD-SAS/mpqp/pull/198/commits/e8b7d912640134196d0c0d18286dd6f9667b8760

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -484,47 +481,8 @@ def _check_targets_order(self):
...
-            return
-
-        if self.nb_qubits != self.observables[0].nb_qubits:
-            raise NumberQubitsError(
-                f"Target size {self.nb_qubits} doesn't match observable size "
-                f"{self.observables[0].nb_qubits}."
-            )
-
+        self.rearranged_targets = list(self.targets)
         self._pre_measure: list[Gate] = []
```
Henri-ColibrITD: is this function still needed ?

mpqp/execution/providers/aws.py
```diff
@@ -187,11 +187,14 @@ def run_braket_observable(job: Job):
 
         if job.measure.pre_transpiled is None:
             grouping = job.measure.get_pauli_grouping()
+            pre_measure = [
+                QCircuit(find_qubitwise_rotations(group)) for group in grouping
+            ]
+            for circuit in pre_measure:
+                for instr in circuit.instructions:
+                    instr.targets[0] = job.measure.targets[instr.targets[0]]
```
Henri-ColibrITD: why are you changing just the first target ? the logic looks a bit dodgy here

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -484,47 +481,8 @@ def _check_targets_order(self):
...
-            self._pre_measure: list[Gate] = []
-            return
-
-        if self.nb_qubits != self.observables[0].nb_qubits:
-            raise NumberQubitsError(
-                f"Target size {self.nb_qubits} doesn't match observable size "
-                f"{self.observables[0].nb_qubits}."
-            )
-
+        self.rearranged_targets = list(self.targets)
```
Henri-ColibrITD: I guess the idea here is than now you support arbitrary order, so all the logic
that was here before is not needed anymore, can you confirm ?

tests/core/instruction/gates/test_custom_gate.py
```diff
@@ -64,7 +64,10 @@ def exec_random_orthogonal_matrix(circ_size: int, device: AvailableDevice):
     result = run(c, device)
 
     # we reduce the precision because of approximation errors coming from CustomGate usage
-    assert isinstance(result, Result)
+    from mpqp.tools import pprint
+
+    pprint(result.amplitudes)
```
JulienCalistoTD: Did you leave a print?

mpqp/execution/providers/aws.py
JulienCalistoTD: why targets[0] ?

mpqp/tools/maths.py
```diff
@@ -478,3 +479,135 @@ def is_power_of_two(n: int) -> bool:
...
+            shuffled[targets[index]],
+            shuffled[index],
+        )
+
+        i = targets.index(index)
+        targets[i], targets[index] = (
+            targets[index],
+            targets[i],
+        )
+    return pauli
```
Henri-ColibrITD: I would move this to be a method of `PauliString`

mpqp/tools/maths.py
```diff
@@ -478,3 +479,135 @@ def is_power_of_two(n: int) -> bool:
...
+    """
+    if copy:
+        from copy import deepcopy
+
+        pauli = deepcopy(ps)
+    else:
+        pauli = ps
+
+    l = len(targets)
+    shuffled = sorted(targets)
```
Henri-ColibrITD: "shuffled" means "mélangé", "rearranged" or "reordered" might be clearer

mpqp/core/instruction/measurement/expectation_value.py
MathieuG-Colibri: Yes I confirm your suspicions.

mpqp/core/instruction/measurement/expectation_value.py
MoHermes: Done 59fd820
Removed `rearranged_targets`, it was only used by the old swap path. Also removed `pre_measure` since `Measure.pre_measure` already returns `[ ]`

mpqp/execution/providers/aws.py
MoHermes: Yeah, `instr.targets[0]` was originally used because `find_qubitwise_rotations(...)` return single qubit instructions., so each generated instruction has one target. But yeah, this assumption is not obvious, so I agree it looks fragile

MoHermes: Done 9ddb86b

tests/core/instruction/gates/test_custom_gate.py
MoHermes: Here 6b12bcc

mpqp/execution/providers/aws.py
MoHermes: https://github.com/ColibrITD-SAS/mpqp/pull/198#discussion_r3317143941

MoHermes: Done 9ddb86b

mpqp/tools/maths.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/198/commits/01a2724199563b6bfebacd22d6bd78942bc1425d

mpqp/tools/maths.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/198/commits/01a2724199563b6bfebacd22d6bd78942bc1425d

mpqp/tools/circuit.py
```diff
@@ -86,6 +88,14 @@ def random_circuit(
...
     for _ in range(nb_gates):
         qcircuit.add(random_gate(gate_classes, nb_qubits, rng))
+
+    if use_all_qubits:  # used in case we want to test braket
+        gates = np.array(
+            [g for g in NATIVE_GATES if issubclass(g, OneQubitNoParamGate)]
+        )
+
+        for i in range(nb_qubits):
+            qcircuit.add(rng.choice(gates)(i))
```
Henri-ColibrITD, line 98: don't we have a function for that ?

mpqp/core/instruction/measurement/expectation_value.py
Henri-ColibrITD: I wonder if you didn't mixed things up, I was talking about the function @MathieuG-Colibri  removed in 4dfbb16b068a21de406d7b25179854f4d9500b1a

mpqp/tools/circuit.py
MathieuG-Colibri: technically yes but not for the case of onequbitnoparam.
So I rewrote it because it didn't fit this case exactly

mpqp/core/instruction/measurement/expectation_value.py
MoHermes: Ok ok, for your original comment about `_check_targets_order()`, no longer needed . My later change was a cleanup of the remaining swap related fields that left behind

mpqp/execution/runner.py
```diff
@@ -55,56 +56,105 @@
...
+        pauli = obs.pauli_string
+        embedded = PauliString()
+
+        for mono in pauli.monomials:
+            full_register = [pI] * nb_qubits
+
+            for local_idx, target in enumerate(targets):
+                full_register[target] = mono.atoms[local_idx]
+
+            embedded += PauliStringMonomial(mono.coef, full_register)
```
Henri-ColibrITD: as julien suggested on the new gate PR, why not else here instead of continue ?

mpqp/tools/maths.py
```diff
@@ -4,17 +4,20 @@
...
 import math
+from copy import copy
 from functools import reduce
 from numbers import Complex, Real
 from typing import TYPE_CHECKING, Any, Optional, Union
 
 import numpy as np
 import numpy.typing as npt
 from scipy.linalg import inv, sqrtm
+from typeguard import typechecked
```
Henri-ColibrITD: I think for optimization reasons we stopped using typeguard (I'm sad but I fold to the god of optimization), is there a reason to reintroduce it ?

mpqp/tools/maths.py
```diff
@@ -478,3 +479,83 @@ def is_power_of_two(n: int) -> bool:
...
+    >>> pprint(matrix)
+    [[1, 0 , 0, 0 ],
+     [0, -1, 0, 0 ],
+     [0, 0 , 1, 0 ],
+     [0, 0 , 0, -1]]
+    >>> pprint(rearrange_matrix(matrix, [1,0]))
+    [[1, 0, 0 , 0 ],
+     [0, 1, 0 , 0 ],
+     [0, 0, -1, 0 ],
+     [0, 0, 0 , -1]]
```
Henri-ColibrITD: since matrices don't need to be unitary here, I would even write matrix as `matrix = np.diag([1,2,3,4])` to make it clearer regarding what's happening

Henri-ColibrITD: also saves you two lines trololo

Henri-ColibrITD (changes requested):
"Une toute petite question" and I'm good !!! 🥳

mpqp/tools/maths.py
MathieuG-Colibri: I wanted to use a matrix that would represent a circuit so that the shuffling of the resulting matrix represents the shuffling of a circuit. which is the spirit of the feature.

mpqp/tools/maths.py
MathieuG-Colibri: Nope no reason just copy pasted some old code in which we used it without thinking.

MathieuG-Colibri: fixed https://github.com/ColibrITD-SAS/mpqp/pull/198/commits/2dba8b86f492eacddf5841e3f8986127259fda20

mpqp/execution/runner.py
MathieuG-Colibri: changed to else: https://github.com/ColibrITD-SAS/mpqp/pull/198/commits/2dba8b86f492eacddf5841e3f8986127259fda20

mpqp/tools/maths.py
Henri-ColibrITD: yeah, but demonstrating this specific function's working here is more important here than the overall effect of the PR, you're writing the example for a single function, not of the PR as a whole

Henri-ColibrITD: this function is more generic than just an operation on observables, it's an operation on any matrix, as such, showing the effect on a clearer matrix makes more sense to me

MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/198/commits/cd3d5a71e8b89b9a944198e92a23839132c86686

Henri-ColibrITD (approved):
GTG!

MoHermes (approved):

PR #199, opened 2026-05-20
Fix simulated devices

MathieuG-Colibri:
Reworked devices and added is_retired to simulated devices

### DIFF
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -1651,12 +1651,12 @@ class QCircuit:
             if job_type == JobType.STATE_VECTOR:
                 skip_measurements = True
 
-            if any(
-                isinstance(i, tuple(device.incompatible_gate()))
+            if len(device.compatible_gate()) != 0 and any(
+                not isinstance(i, tuple(device.compatible_gate()))
                 for i in self.instructions
             ):
                 raise ValueError(
-                    f"Gate(s) {', '.join(map(str, device.incompatible_gate()))} cannot be simulated on {device}."
+                    f"Gate(s) {', '.join(map(str, device.compatible_gate()))} cannot be simulated on {device}."
                 )
             if (
                 isinstance(device, StaticIBMSimulatedDevice)
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -1657,7 +1657,7 @@ class QCircuit:
                 if any(type(i) not in compatible_gates for i in self.gates):
                     raise ValueError(
                         f"Gates {', '.join(map(str, compatible_gates))} "
-                        f"are the only one available on {device}."
+                        f"are the only ones available on {device}."
                     )
             if (
                 isinstance(device, StaticIBMSimulatedDevice)
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -51,12 +51,6 @@ from mpqp.core.instruction.gates.parametrized_gate import ParametrizedGate
 from mpqp.core.instruction.measurement import BasisMeasure, Measure
 from mpqp.core.instruction.measurement.expectation_value import ExpectationMeasure
 from mpqp.core.languages import Language
-from mpqp.environment.var_cache import (
-    _INSTALLED_MPQP_PROVIDERS,  # pyright: ignore[reportPrivateUsage]
-)
-from mpqp.environment.var_cache import (
-    InstalledProviders,
-)
 from mpqp.noise.noise_model import DimensionalNoiseModel, NoiseModel
 from mpqp.tools.errors import (
     DeviceJobIncompatibleError,
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -1652,10 +1652,7 @@ class QCircuit:
                 skip_measurements = True
             gate_set = list(device.compatible_gate())
             if len(gate_set) != 0:
-                if any(
-                    type(i) not in gate_set
-                    for i in self.without_measurements().instructions
-                ):
+                if any(type(i) not in gate_set for i in self.gates):
                     raise ValueError(
                         f"Gate(s) {', '.join(map(str, device.compatible_gate()))} cannot be simulated on {device}."
                     )
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -53,15 +53,17 @@ from mpqp.core.instruction.measurement.expectation_value import ExpectationMeasu
 from mpqp.core.languages import Language
 from mpqp.environment.var_cache import (
     _INSTALLED_MPQP_PROVIDERS,  # pyright: ignore[reportPrivateUsage]
+)
+from mpqp.environment.var_cache import (
     InstalledProviders,
 )
 from mpqp.noise.noise_model import DimensionalNoiseModel, NoiseModel
 from mpqp.tools.errors import (
     DeviceJobIncompatibleError,
+    InstructionAfterMeasurementError,
     InstructionParsingError,
     NonReversibleWarning,
     NumberQubitsError,
-    InstructionAfterMeasurementError,
 )
 from mpqp.tools.generics import OneOrMany
 from mpqp.tools.maths import matrix_eq
@@ -1650,11 +1652,12 @@ class QCircuit:
         if isinstance(device, (IBMDevice, StaticIBMSimulatedDevice)):
             if job_type == JobType.STATE_VECTOR:
                 skip_measurements = True
-            gate_set = list(device.compatible_gate())
-            if len(gate_set) != 0:
-                if any(type(i) not in gate_set for i in self.gates):
+            compatible_gates = list(device.compatible_gates())
+            if len(compatible_gates) != 0:
+                if any(type(i) not in compatible_gates for i in self.gates):
                     raise ValueError(
-                        f"Gate(s) {', '.join(map(str, device.compatible_gate()))} cannot be simulated on {device}."
+                        f"Gate(s) {', '.join(map(str, compatible_gates))} "
+                        f"cannot be simulated on {device}."
                     )
             if (
                 isinstance(device, StaticIBMSimulatedDevice)
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -1656,8 +1656,8 @@ class QCircuit:
             if len(compatible_gates) != 0:
                 if any(type(i) not in compatible_gates for i in self.gates):
                     raise ValueError(
-                        f"Gate(s) {', '.join(map(str, compatible_gates))} "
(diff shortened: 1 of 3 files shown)


mpqp/core/circuit.py
```diff
@@ -1650,14 +1650,15 @@ def to_other_device(
...
                 skip_measurements = True
-
-            if any(
-                isinstance(i, tuple(device.incompatible_gate()))
-                for i in self.instructions
-            ):
-                raise ValueError(
-                    f"Gate(s) {', '.join(map(str, device.incompatible_gate()))} cannot be simulated on {device}."
-                )
+            gate_set = list(device.compatible_gate())
```
hJaffaliColibritd: What does it mean that a device doesn't have a compatible gate set ? 
Isn't it equivalent to say `is_remote()` or something else ?

mpqp/core/circuit.py
```diff
@@ -1650,14 +1650,15 @@ def to_other_device(
...
+            if len(gate_set) != 0:
+                if any(
+                    type(i) not in gate_set
+                    for i in self.without_measurements().instructions
```
hJaffaliColibritd: Why not using `self.gates()` ?

mpqp/core/circuit.py
MathieuG-Colibri: So I changed the check from -> excluding all unauthorized gates to specifiy all the gates we can use.

It still checks it below

mpqp/core/circuit.py
MathieuG-Colibri: Forgot it was an option added it https://github.com/ColibrITD-SAS/mpqp/pull/199/commits/c09dfea55f29f6a3218b6aae7cbc1c0fff2e6aa9

mpqp/core/circuit.py
MathieuG-Colibri: Nevermind missunderstood the comment.
This method should be called if the user doesn't want any transpilation (verbatim for example) and when the gate isn't supported by the device.

mpqp/execution/devices.py
```diff
@@ -90,7 +92,13 @@ def supports_observable(self) -> bool:
     def supports_observable_ideal(self) -> bool:
         pass
 
-    def incompatible_gate(self) -> set[type[Gate]]:
+    def compatible_gate(self, native_set: bool = False) -> set[type[Gate]]:
```
hJaffaliColibritd: This means that each time we add a gate to MPQP (which is not that often I admit) we have to remember to update each time all the compatible gates ?

mpqp/execution/devices.py
```diff
@@ -512,3 +525,14 @@ def supports_observable(self) -> bool:
...
+
+IBM_CHIPS_HERON = [IBMDevice.IBM_MIAMI, IBMDevice.IBM_BERLIN]
+IBM_CHIPS_NIGHTHAWK = [
+    IBMDevice.IBM_BOSTON,
+    IBMDevice.IBM_KINGSTON,
+    IBMDevice.IBM_PITTSBURGH,
+    IBMDevice.IBM_FEZ,
+    IBMDevice.IBM_MARRAKESH,
+    IBMDevice.IBM_AACHEN,
+]
```
hJaffaliColibritd: A bit original to do it like this compared to the rest of devices?
Maybe a method to return that?

hJaffaliColibritd (approved):

mpqp/execution/devices.py
MathieuG-Colibri: We could have a method but chips family like is more often done by IBM.
It felt bad adding a whole method for AvailableDevice only for an edge case of qiskit.

mpqp/execution/devices.py
MathieuG-Colibri: No because more often than not we're going to transpile the circuit so we won't have forbidden gates by themselves.
For the cases that it's really needed we can do a list product with the NATIVE_GATES constant.

mpqp/core/circuit.py
hJaffaliColibritd: Ok

mpqp/execution/devices.py
hJaffaliColibritd: The idea was to avoid instantiating a list the first time we import the file, but here it is not that a big deal

mpqp/core/circuit.py
MoHermes: I agree with native/verbatim validation. I think we just guard it ? so `device.compatible_gate()` is only used in that path, instead of always validating the original instructions unconditionally before transpilation.

mpqp/core/circuit.py
```diff
@@ -1650,14 +1650,12 @@ def to_other_device(
...
-                for i in self.instructions
-            ):
-                raise ValueError(
-                    f"Gate(s) {', '.join(map(str, device.incompatible_gate()))} cannot be simulated on {device}."
-                )
+            gate_set = list(device.compatible_gate())
+            if len(gate_set) != 0:
+                if any(type(i) not in gate_set for i in self.gates):
+                    raise ValueError(
+                        f"Gate(s) {', '.join(map(str, device.compatible_gate()))} cannot be simulated on {device}."
```
Henri-ColibrITD: was it more efficient to reverse the logic ? (I'm genuinely interested)

mpqp/core/circuit.py
```diff
@@ -1650,14 +1652,13 @@ def to_other_device(
...
-                for i in self.instructions
-            ):
-                raise ValueError(
-                    f"Gate(s) {', '.join(map(str, device.incompatible_gate()))} cannot be simulated on {device}."
-                )
+            compatible_gates = list(device.compatible_gates())
+            if len(compatible_gates) != 0:
+                if any(type(i) not in compatible_gates for i in self.gates):
+                    raise ValueError(
+                        f"Gate(s) {', '.join(map(str, compatible_gates))} "
```
Henri-ColibrITD: is there a logic error here ?

mpqp/core/circuit.py
MathieuG-Colibri: Yes because in some cases (for verbatim for example) since we don't want transpilation we need the gate set of the hardware. Since the hardware often doesn't support many gates (often 2-3) it's faster to check and to write a sample of authorized gates rather than the inverse.

Also better for future proofing when we'll have more gates.

mpqp/execution/devices.py
```diff
@@ -111,29 +120,23 @@ class IBMDevice(AvailableDevice):
...
-    IBM_FEZ = "ibm_fez"
     IBM_RENSSELAER = "ibm_rensselaer"
-    IBM_BRUSSELS = "ibm_brussels"
     IBM_KAWASAKI = "ibm_kawasaki"
     IBM_QUEBEC = "ibm_quebec"
-    IBM_TORINO = "ibm_torino"
-    IBM_NAZCA = "ibm_nazca"
-    IBM_STRASBOURG = "ibm_strasbourg"
-
-    # RETIRED - IBM_OSAKA = "ibm_osaka"
```
Henri-ColibrITD, line 127: not sure but keeping the retired devices commented might be useful, I don't have a strong opinion though so if you don't like it I'm ok as is

MathieuG-Colibri: I thought it was clogging the file for not much use in the end. But I don't have any strong feelings one way or the other.

mpqp/core/circuit.py
MathieuG-Colibri: If **any** of the gates in the circuit is **not** part of the authorized gate set then we raise an error.

Seems good to me

mpqp/execution/devices.py
Henri-ColibrITD: do we need this list elsewhere ? if it's only for the compatible gates, we could have it in the method only

Henri-ColibrITD: I would even see a nice way to solve it with dictionary, something like a `device_generation: dict[IBMDevice, DeviceGeneration]` and a `compatibility_matrix: dict[DeviceGeneration, set[Gate]]`, this way the compatible gate is simply `compatibility_matrix[device_generation[self]]` (with dict, methods, etc..., and DeviceGeneration being an enum, maybe IBMDeviceGeneration even ?)

mpqp/execution/simulated_devices.py
```diff
@@ -85,6 +85,19 @@ def get_ibm_fake_providers() -> list[tuple[str, type["FakeBackendV2"]]]:
...
             return []
 
+    def is_retired(self) -> bool:
+        """Function used to tell if the simulated device is retired or not.
+        Note: It only compare its name to a currently existing one, if an old simulated device has the same name as a new one it may break.
+        """
+        from mpqp.execution.devices import IBMDevice
+
+        name = self.name[4:]
+        if name[-2:] == "V2":
```
Henri-ColibrITD: str's `endswith` ?

mpqp/execution/simulated_devices.py
```diff
@@ -85,6 +85,19 @@ def get_ibm_fake_providers() -> list[tuple[str, type["FakeBackendV2"]]]:
...
+            name = name[:-2]
+        name = name.upper()
+        devices_names = IBMDevice._member_names_
+        return not any(name in device for device in devices_names)
```
Henri-ColibrITD, line 100: so the source of truth is our `IBMDevice` enum ? there is no check independant of our implem ? like just the fact that it ends with V2, an IBM endpoint or whatever ?

Henri-ColibrITD, line 100: we might be a bit more precise than `name in device` like `"IBM_"+name == device` ?

MathieuG-Colibri: Yes it bothered me also but since we don't generate dynamically our enum we don't have another point of reference. We need to keep it updated anyway.

Henri-ColibrITD (approved):
aller là

mpqp/core/circuit.py
```diff
@@ -1650,14 +1652,13 @@ def to_other_device(
...
-            ):
-                raise ValueError(
-                    f"Gate(s) {', '.join(map(str, device.incompatible_gate()))} cannot be simulated on {device}."
-                )
+            compatible_gates = list(device.compatible_gates())
+            if len(compatible_gates) != 0:
+                if any(type(i) not in compatible_gates for i in self.gates):
+                    raise ValueError(
+                        f"Gates {', '.join(map(str, compatible_gates))} "
+                        f"are the only one available on {device}."
```
Henri-ColibrITD:
```suggestion
                        f"are the only ones available on {device}."
```

mpqp/core/circuit.py
Henri-ColibrITD: nop, but we modified it together so all good for me

MoHermes (approved):

PR #200, opened 2026-05-20
Feat cirq custom gate

MathieuG-Colibri:
Change the cirq custom gate implem and behavior
Cleanup and bug fixing

Refactored to_other_language


mpqp/core/instruction/gates/custom_controlled_gate.py
```diff
@@ -34,11 +34,13 @@ class CustomControlledGate(ControlledGate):
...
-        q_1: ┤ Unitary ├
-             └────┬────┘
-        q_2: ─────■─────
+             ┌──────────┐
+        q_0: ┤2         ├
+             │          │
+        q_1: ┤1 Unitary ├
+             │          │
+        q_2: ┤0         ├
+             └──────────┘
```
Henri-ColibrITD: this looks like a regression to me: previous version seemed more representative

mpqp/tools/cirq.py
```diff
@@ -0,0 +1,47 @@
+from typing import Any, Optional
+
+from mpqp.environment.var_cache import (
+    _INSTALLED_MPQP_PROVIDERS,  # pyright: ignore[reportPrivateUsage]
```
Henri-ColibrITD: is there a reason to have it private ? I don't remember

mpqp/tools/unitary_decomposition.py
```diff
@@ -62,13 +62,18 @@ def _unitary_SVD(U: Matrix) -> tuple[Matrix, Matrix, Matrix]:
...
 
     # Build G as G = V @ D² @ V†
-    G = G0 @ G1.conj().T
+    g = G0 @ G1.conj().T
+    G = np.round(g, 10)
+    eigvals, v = np.linalg.eig(G)
+    from mpqp.tools import closest_unitary, is_unitary
+
+    if not is_unitary(v):
+        v = closest_unitary(v)
```
Henri-ColibrITD: we could move this test in "clostest_unitary"

mpqp/tools/unitary_decomposition.py
```diff
@@ -118,14 +124,16 @@ def _gray_code_decomposition(
         # CNOT's control is the changed bit of two consecutive numbers in gray code
         changed = _gray_code(i) ^ _gray_code(i + 1)
         control = next(i for i in range(len(thetas)) if (changed >> i & 1))
-        control = max(-control - position - 1 + circuit.nb_qubits, 1)
+        control = max(-control - targets[position] - 1 + circuit.nb_qubits, 1)
         if np.abs(angle) > PRECISION:  # Dodge unnecessary rotations
-            circuit.add(rotation(angle, position))
-        circuit.add(CNOT(control + position, position))
+            circuit.add(rotation(angle, int(targets[position])))
+        circuit.add(CNOT(int(control + targets[position]), int(targets[position])))
```
Henri-ColibrITD: why int casting ?

mpqp/tools/unitary_decomposition.py
Henri-ColibrITD: not part of the PR but not super clear

mpqp/tools/unitary_decomposition.py
```diff
@@ -118,14 +124,16 @@ def _gray_code_decomposition(
         # CNOT's control is the changed bit of two consecutive numbers in gray code
         changed = _gray_code(i) ^ _gray_code(i + 1)
         control = next(i for i in range(len(thetas)) if (changed >> i & 1))
-        control = max(-control - position - 1 + circuit.nb_qubits, 1)
+        control = max(-control - targets[position] - 1 + circuit.nb_qubits, 1)
```
Henri-ColibrITD, line 128: linked to previous comment

mpqp/tools/unitary_decomposition.py
```diff
@@ -140,13 +148,17 @@ def _decompose(U: Matrix, circuit: QCircuit, position: int = 0) -> QCircuit:
 
         # extract the global phase so that SU is a special unitary, note that if U is a special unitary then delta = 0
         SU = U / np.exp(1j * delta)
-        beta = 2 * math.acos(np.abs(SU[0][0]))
+        abs = np.abs(SU[0][0])
+        if abs >= 1:
+            beta = 0
+        else:
+            beta = 2 * math.acos(abs)
```
Henri-ColibrITD: small preference for ternary here but you can ignore this comment if you will

mpqp/tools/unitary_decomposition.py
```diff
@@ -140,13 +148,17 @@ def _decompose(U: Matrix, circuit: QCircuit, position: int = 0) -> QCircuit:
...
         alpha = -np.angle(SU[0][0]) - np.angle(SU[1][0])
         gamma = -np.angle(SU[0][0]) + np.angle(SU[1][0])
 
-        circuit.add(Rz(alpha, position))
-        circuit.add(Ry(beta, position))
-        circuit.add(Rz(gamma, position))
+        circuit.add(Rz(alpha, int(targets[position])))
+        circuit.add(Ry(beta, int(targets[position])))
+        circuit.add(Rz(gamma, int(targets[position])))
```
Henri-ColibrITD: same as before, why the int casting ? targets is supposed to be a list of ints

tests/execution/test_validity.py
```diff
@@ -1086,3 +1086,29 @@ def test_validity_optim_ideal_multi_diag_obs_and_regular_run(
...
+        IBMDevice.AER_SIMULATOR,
+        GOOGLEDevice.CIRQ_LOCAL_SIMULATOR,
+        AWSDevice.BRAKET_LOCAL_SIMULATOR,
+        ATOSDevice.MYQLM_PYLINALG,
+    ]
+    results = run(circuit, devices)
+    for i in range(len(results) - 1):
+        assert matrix_eq(
+            results[i].state_vector.vector, results[i + 1].state_vector.vector
+        )
```
Henri-ColibrITD:
```suggestion
    for result_1, result_2 in pairwise(results):
        assert matrix_eq(result_1.state_vector.vector, results_2.state_vector.vector)
```

tests/execution/test_validity.py
```diff
@@ -1086,3 +1086,29 @@ def test_validity_optim_ideal_multi_diag_obs_and_regular_run(
...
+@pytest.mark.parametrize(
+    "matrix, gphase",
+    [
+        (rand_unitary_matrix(2), 2),
+        (rand_unitary_matrix(4), -1),
+        (rand_unitary_matrix(8), 10),
+    ],
+)
+def test_global_phase_statevector(matrix: Matrix, gphase: float):
+    from math import log2
```
Henri-ColibrITD, line 1100:
```suggestion
    from math import log2
    from itertools import pairwise
```

mpqp/tools/cirq.py
```diff
@@ -0,0 +1,47 @@
...
+    InstalledProviders,
+)
+
+if InstalledProviders.BRAKET in _INSTALLED_MPQP_PROVIDERS:
```
Henri-ColibrITD: braket ? cirq ?

mpqp/core/instruction/gates/custom_controlled_gate.py
MathieuG-Colibri: Yes but it made that we couldn't translate them back (cannot decompose into qasm).

So it had translatability issues.

mpqp/tools/cirq.py
MathieuG-Colibri: I don't know either I just know it was done that way before.

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: It was a bug with a test where the typing got messed up here and it broken when adding gates.
So it's a failsafe that I put here because I never saw it happen anywhere else.

mpqp/core/circuit.py
Henri-ColibrITD, line 1: moving the export to another file, why not, but maybe we should give it some more thought ? because as is we are  duplicating some code for each supported language

mpqp/core/circuit.py
```diff
@@ -1968,13 +1756,57 @@ def from_other_language(
...
+                    cir, phase = parse_qasm2_gates(
+                        qcircuit.from_moments(qcircuit[split_index:]).to_qasm()
+                    )
+                    c += qasm2_parse(cir)
+                    c.input_g_phase += phase
+                from mpqp.gates import Id
+
+                # This method adds a lot of identities so for a much cleaner circuit we remove them.
+                # Maybe add this as an option
+                c.instructions = [i for i in c.instructions if not isinstance(i, Id)]
```
Henri-ColibrITD: what are we doing with c ? it seems to me like we completely ignore it once built

mpqp/tools/circuit.py
```diff
@@ -353,3 +364,468 @@ def replace_custom_gate(
...
+                f"The gate {type(gate)} are not in the set of authorized gates: f{authorized_gates}"
+            )
+        else:
+            return [gate]
+    if isinstance(gate, CustomControlledGate) and isinstance(
+        gate.non_controlled_gate, CustomGate
+    ):
+        # If the CustomControlledGate contains itself a custom gate it's better to return a bigger custom gate for compatibilities issues.
+        # If the non_controlled_gate is a NativeGate it shouldn't pose a problem.
+        # TODO: check how to decompose a ComposeGate that is inside a CCG, (example: a C-PRX)
```
Henri-ColibrITD: maybe have an exception for circuit drawing ? to solve the first comment I wrote in the PR

mpqp/tools/circuit.py
```diff
@@ -353,3 +364,468 @@ def replace_custom_gate(
...
+        return [gate.to_custom_gate()]
+
+    return [gate]
+
+
+def mpqp_to_qiskit(
```
Henri-ColibrITD: a inconvenient of moving the code here is that it makes the job of reviewing much trickier

Henri-ColibrITD: could we have a meeting to recap what changed ? it would save me some time ^^

Henri-ColibrITD (changes requested):

mpqp/core/instruction/gates/custom_controlled_gate.py
Henri-ColibrITD: there is a difference between display and what's actually stored, see this comment: https://github.com/ColibrITD-SAS/mpqp/pull/200#discussion_r3442435197

mpqp/tools/unitary_decomposition.py
Henri-ColibrITD: maybe we should rather fix the test ?

mpqp/core/circuit.py
MathieuG-Colibri: My main reason for this change is that in some translation context we might need provider specific parameters that would be tougher to keep clean in one to_other_language function.

mpqp/tools/circuit.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/df88af6904bef245cc8a547c9482cf04d9a3828b

mpqp/tools/cirq.py
MathieuG-Colibri: fixed https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/df88af6904bef245cc8a547c9482cf04d9a3828b

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: removed all int casting https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/df88af6904bef245cc8a547c9482cf04d9a3828b

The issue seems to be fixed

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/df88af6904bef245cc8a547c9482cf04d9a3828b

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/df88af6904bef245cc8a547c9482cf04d9a3828b

mpqp/core/instruction/gates/custom_controlled_gate.py
MathieuG-Colibri: done with print option https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/df88af6904bef245cc8a547c9482cf04d9a3828b

mpqp/core/circuit.py
MathieuG-Colibri: refactored https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/df88af6904bef245cc8a547c9482cf04d9a3828b

mpqp/tools/circuit.py
```diff
@@ -353,3 +364,457 @@ def replace_custom_gate(
...
+    circuit: QCircuit,
+    skip_pre_measure: bool = False,
+    skip_measurements: bool = False,
+    authorized_gates: set[type[Gate]] | None = None,
+) -> cirq_Circuit:
+    """Translate a MPQP circuit to a Cirq equivalent.
+
+    Note:
+        If the circuit contains ComposedGate type instructions you need to include the gate in the authorize_gates set for it to avoid decomposition, otherwise, the translation will always
+        try to use the gate's decomposition (see example). Also in the case of CustomGates it will always be decomposed into the {Ry, Rz, CNOT} basis for translation compatibilities.
```
MoHermes: To update docs, we no longer decomposed into the {Ry, Rz, CNOT}

mpqp/tools/circuit.py
```diff
@@ -353,3 +364,457 @@ def replace_custom_gate(
...
+    else:
+        new_circ = QuantumCircuit(circuit.nb_qubits, circuit.nb_cbits)
+
+    if circuit.label is not None:
+        new_circ.name = circuit.label
+
+    for instruction in circuit.instructions:
+        if isinstance(instruction, (Measure, Breakpoint)):
+            continue
+        options = {"printing": printing} if isinstance(instruction, CustomGate) else {}
```
MoHermes: So, ``printing=True`` apply only when the instruction is CustomGate?, I mean it is also propagated CustomControlledGate(...)?

mpqp/core/instruction/gates/custom_controlled_gate.py
```diff
@@ -88,10 +88,11 @@ def to_other_language(
         if language == Language.QISKIT:
             from qiskit.quantum_info import Operator
 
-            gate = self.non_controlled_gate.to_other_language()
+            gate = self.non_controlled_gate.to_other_language(Language.QISKIT)
             if isinstance(gate, Operator):
                 gate = gate.to_instruction()
             gate = gate.control(len(self.controls))
+            gate.control()
```
MoHermes: Is this call intentional?

MoHermes (changes requested):

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -127,12 +127,11 @@ def to_other_language(
...
                 dummy_circuit = QuantumCircuit(self.nb_qubits)
-                for param in qiskit_parameters:
-                    # Rx is just a random choice so to have the parameter in the
-                    # list of inputs
-                    dummy_circuit.rx(param, 0)
-                return dummy_circuit.to_gate(label="CustomGate")
-            return UnitaryGate(self.matrix)
+                dummy_circuit.id(0)
+                return dummy_circuit.to_gate(
+                    label=f"CustomGate({', '.join([str(s) for s in gate_symbols])})"
```
hJaffaliColibritd, line 132: Why did we change to this? the previous way of generating the label was not taking into account all the gate symbols?

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -249,7 +225,22 @@ def decompose(self) -> "QCircuit":
         """
         from mpqp.tools.unitary_decomposition import quantum_shannon_decomposition
 
-        return quantum_shannon_decomposition(self.matrix)
+        # non ordered targets
+        if any(
```
hJaffaliColibritd, line 229: raise a warning or something in this case?

mpqp/tools/circuit.py
MathieuG-Colibri: Solved (forgot which commit)

mpqp/tools/circuit.py
MathieuG-Colibri: Yes printing was used at the beginning to make sure we could print parametrized CG, I did something similar with Custom Controlled Gates.

mpqp/core/instruction/gates/custom_gate.py
MathieuG-Colibri: Yes I think the issue was that not all the params were displayed.

mpqp/core/instruction/gates/custom_gate.py
MathieuG-Colibri: Is it necessary to raise a warning if the edge case is taken into account anyway ?

mpqp/core/instruction/gates/custom_controlled_gate.py
MathieuG-Colibri: Removed https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/330e5800fa990b6029f246e9564bee36696cf2c0

mpqp/tools/circuit.py
hJaffaliColibritd: Maybe this was already discussed but I would like to know the reason why we would keep this here? (I take liberty to unresolve the comment)
I meen mpqp.tools.circuit is supposed to be for tools on manipulation of circuit. Anything related to translation should have another dedicated file/location in the code I believe

hJaffaliColibritd: @MathieuG-Colibri so my suggestion, as discussed this week, is to regroup all translation/conversion features and code in a new module/package `mpqp.translation`or whatever.
The idea is to regroup all specifc functions/block of code called in `to_other_language` and `from_other_language`, or other helper functions or tools related to translation. This should also include the package concerning qasm conversion, (we would typically now have mpqp.translation.qasm instead of mpqp.qasm for instance)

mpqp/core/instruction/gates/custom_gate.py
hJaffaliColibritd: Is there a risk, when we have too much parameters, that the display become weird ? in that case we can put "..." and not list them all but the first and last.
If this never happens you can resolve this comment

mpqp/core/instruction/gates/custom_gate.py
hJaffaliColibritd: It's taken into account by tweaking the user's input, which may affect few stuff between what he submitted and what is actually run on the backend, right?
In that case it is nice if we can inform, or warn the user of the changes we make.

mpqp/tools/maths.py
```diff
@@ -197,6 +197,8 @@ def closest_unitary(matrix: Matrix) -> Matrix:
         True
 
     """
+    if is_unitary(matrix):
```
hJaffaliColibritd: This could slow down the performance of the function, if we need to check for unitary each time we call this.
I think it is better to suppose that if the user calls this function, it means he already knows that the matrix is not unitary and wants to get the closer unitary.
we can precise that in the documentation of the function.

mpqp/core/instruction/gates/custom_gate.py
```diff
@@ -249,7 +225,22 @@ def decompose(self) -> "QCircuit":
...
+        ):
+            from copy import deepcopy
+
+            from mpqp.tools import rearrange_matrix
+
+            targets = deepcopy(self.targets)
+            targets.sort()
+
+            return quantum_shannon_decomposition(
+                rearrange_matrix(self.matrix, self.targets, do_copy=True), targets
```
hJaffaliColibritd, line 244: I checked the tests of `rearrange_matrix`, it looks like they are not really independent from the specific implementation we made, since it reuses it behind the usage of `CustomGate`, I would add some specific "manual" examples to be 100% sure.

mpqp/core/circuit.py
hJaffaliColibritd: This goes with another comment somewhere else, leaning towards creating mpqp.translation : https://github.com/ColibrITD-SAS/mpqp/pull/200/changes#r3442442852

I resolve this comment and we can move the discussion to the other comment

mpqp/core/circuit.py
```diff
@@ -1174,6 +1180,7 @@ def to_other_language(
         language: Language = Language.QISKIT,
         skip_pre_measure: bool = False,
         skip_measurements: bool = False,
+        authorized_gates: Optional[set[type[Gate]]] = None,
```
hJaffaliColibritd: 1. Explanation of the argument missing in the doc
2. Why not `Optional[set[type[NativeGate]]]` instead?
Using `Gate` allows for custom implementation of the Abstract class `Gate`, which may not be in the list of nativegate we handle in the library, which can cause problem at the translation ?

mpqp/tools/circuit.py
MathieuG-Colibri: Done ! https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/ef5dcdabf1b447a7c3e6eb52767663cb9875cc98

mpqp/execution/providers/google.py
```diff
@@ -610,5 +610,9 @@ def extract_result_STATE_VECTOR(
     Returns:
         The formatted result.
     """
-    state_vector = StateVector(result.final_state_vector, job.circuit.nb_qubits)
-    return Result(job, state_vector, 0, 0)
+    state_vector = result.final_state_vector
+    if job.circuit.input_g_phase:
+        import numpy as np
+
+        state_vector = state_vector * np.exp(-1j * job.circuit.input_g_phase)
```
hJaffaliColibritd: This is normally directly done in `Result` no?

MathieuG-Colibri: Not in the case of cirq (couldn't have known because we didn't implement the cirq gphase before)

mpqp/core/circuit.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/9e33b6cafb275875860702c92504c8f26e5b56ec

mpqp/core/instruction/gates/custom_gate.py
MathieuG-Colibri: Actually Henri wanted the example to be not circuit related because it's in tools/math.

Nevertheless done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/9e33b6cafb275875860702c92504c8f26e5b56ec

mpqp/tools/maths.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/9e33b6cafb275875860702c92504c8f26e5b56ec

mpqp/core/instruction/gates/custom_gate.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/9e33b6cafb275875860702c92504c8f26e5b56ec

docs/translation.rst
```diff
@@ -0,0 +1,89 @@
...
+    This module is separated in two parts:
+
+    - QASM converter
+    - Direct MPQP to provider and provider to MPQP translations
+
+.. code-block:: python
+    :class: import
+
+    from mpqp.translation import *
+
```
hJaffaliColibritd:
```suggestion

.. code-block:: python
    :class: import

    from mpqp.translation import *
    
This module regroups all the modules and functions used to import quantum circuits and other objects from providers' SDKs, but also to export MPQP objects.

This module contains:
    - A submodule regrouping all features allowing interoperability through OpenQASM
    - Provider-specific functions used for import and export of quantum objects (circuits, observables, noise models, etc.)

```

mpqp/tools/cirq.py
```diff
@@ -0,0 +1,47 @@
+from typing import Any, Optional
```
hJaffaliColibritd: I guess this file should be move to mpqp.translation

mpqp/tools/maths.py
```diff
@@ -176,12 +176,13 @@ def is_diagonal(matrix: npt.NDArray[Any]) -> bool:
...
 
 
-def closest_unitary(matrix: Matrix) -> Matrix:
+def closest_unitary(matrix: Matrix, check_unitary: bool = False) -> Matrix:
     """Calculate the unitary matrix that is closest with respect to the operator
     norm distance to the general matrix in parameter.
 
     Args:
         matrix: Matrix for which we want to determine the closest unitary matrix.
+        check_unitary: If true checks if the matrix is unitary. If the matrix is unitary return it.
```
hJaffaliColibritd:
```suggestion
        check_unitary: If True, we check whether the input matrix is unitary, and return it if so.
```

mpqp/tools/unitary_decomposition.py
```diff
@@ -62,13 +62,17 @@ def _unitary_SVD(U: Matrix) -> tuple[Matrix, Matrix, Matrix]:
     G1 = np.array(g1)
 
     # Build G as G = V @ D² @ V†
-    G = G0 @ G1.conj().T
+    g = G0 @ G1.conj().T
+    G = np.round(g, 10)
+    eigvals, v = np.linalg.eig(G)
```
hJaffaliColibritd, line 68: So we compute the eigenvalues and vectors on rounded (potentially non-unitary) matrix G, before closest unitary ?

Also here in line 67 you compute eigenvectors of G that you store in `v`, but then you overrite `v`to put the closest unitary of itself, which means that nows the eigvals may not be exactly the ones of `eigvals`, can this cause a problem?

mpqp/tools/unitary_decomposition.py
```diff
@@ -249,10 +266,18 @@ def quantum_shannon_decomposition(U: Matrix) -> QCircuit:
...
+            f"The size of the unitary matrix should be of a power of 2: got {len(U)}"
         )
-    circuit = QCircuit(int(np.log2(len(U))))
-    circuit = _decompose(U, circuit, 0)
+    import numpy as np
+
+    if targets is None:
+        targets = list(range(np.log2(len(U)).astype(int)))
+    circuit = QCircuit()
+    circuit = _decompose(U, circuit, targets, 0)
```
hJaffaliColibritd:
```suggestion
    circuit = _decompose(U, QCircuit(), targets, 0)
```

mpqp/translation/braket.py
```diff
@@ -0,0 +1,181 @@
...
+from mpqp.core.instruction.gates.custom_controlled_gate import CustomControlledGate
+from mpqp.core.instruction.gates.custom_gate import CustomGate
+from mpqp.core.instruction.gates.gate import Gate
+from mpqp.core.instruction.gates.native_gates import CRk, NativeGate
+from mpqp.core.languages import Language
+from mpqp.environment.var_cache import (
+    _INSTALLED_MPQP_PROVIDERS,  # pyright: ignore[reportPrivateUsage]
+    InstalledProviders,
+)
+from mpqp.translation.utils import verify_convert_instructions
```
hJaffaliColibritd: Can we optimize imports here?
Import classes for type checking, or functions etc closer to the place they are called for instance.

mpqp/translation/utils.py
```diff
@@ -0,0 +1,30 @@
+from mpqp.core.instruction.gates.custom_controlled_gate import CustomControlledGate
+from mpqp.core.instruction.gates.custom_gate import CustomGate
+from mpqp.core.instruction.gates.gate import Gate
+from mpqp.core.instruction.gates.native_gates import NativeGate
```
hJaffaliColibritd: Can we optimize imports here?
Import classes for type checking, or functions etc closer to the place they are called for instance. 
Or if it is used for the typechecker then put them inside `if TYPE_CHECKING`

mpqp/core/circuit.py
```diff
@@ -1850,6 +1638,7 @@ def to_other_device(
     def from_other_language(
         cls,
         qcircuit: QuantumCircuit | cirq_Circuit | braket_Circuit | myQLM_Circuit | str,
+        ignore_identities: bool = True,
```
hJaffaliColibritd: Add description in the doc

hJaffaliColibritd (changes requested):

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: @MathieuG-Colibri can you give an explanation why we needed to introduce targets[position]?

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: Up to e-10 it doesn't make much of a difference, it's ugly and arbitrary but it works well still.

mpqp/tools/cirq.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/07d2af574bd7bb3499598164b76f6612350e07f0

mpqp/translation/utils.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/07d2af574bd7bb3499598164b76f6612350e07f0

mpqp/translation/braket.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/07d2af574bd7bb3499598164b76f6612350e07f0

mpqp/core/circuit.py
MathieuG-Colibri: removed it was unused now https://github.com/ColibrITD-SAS/mpqp/pull/200/commits/07d2af574bd7bb3499598164b76f6612350e07f0

mpqp/execution/providers/google.py
hJaffaliColibritd: But don't we want to do it in the Result class, like we do for other providers, instead of here? 
It would be cleaner if we group this kind of correction to the same place no?

mpqp/tools/unitary_decomposition.py
hJaffaliColibritd: I would put a TODO or a FIXME in the code, because it can potentially numerically cause some problems, especially for larger matrices

mpqp/core/instruction/gates/custom_gate.py
hJaffaliColibritd: I'm not reffering to the example in the documentation, I was more talking about tests, in the `tests` folder. If you look at them, we reuse CustomGate in the test, which makes the comparison in the tests not independant from the feature.

mpqp/tools/unitary_decomposition.py
MathieuG-Colibri: It was in the context of custom gate if the custom gate is in a circuit and the position starts at [1,2,3] instead of [0,1,2] if needed.

It doesn't change the result in case of the decomposition of a sole matrix.

PR #204, opened 2026-07-10
Feat: Statevector to sample list

joeffreyBelot:
I was trying to use StateVector to do some computations and analysis, but it was not convenient to each time having to parse the numpy array to deduce what are the basis states.
I saw you had Samples, so I added a method to get a list of samples so I can look at their binary representation directly

### DIFF
--- a/mpqp/execution/result.py
+++ b/mpqp/execution/result.py
@@ -125,7 +125,11 @@ class StateVector:
             [Sample(2, index=0, count=None, probability=0.5), Sample(2, index=3, count=None, probability=0.5)]
 
         """
-        pass  
+        return [
+            Sample(self.nb_qubits, index=index, probability=proba)
+            for index, proba in enumerate(self.probabilities)
+            if not np.isclose(proba, 0)
+        ] 
 
     def __eq__(self, other) -> bool:  # pyright: ignore[reportMissingParameterType]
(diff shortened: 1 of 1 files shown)


hJaffaliColibritd:
Changed target branch to dev @joeffreyBelot

mpqp/execution/result.py
```diff
@@ -106,6 +106,31 @@ def to_dict(self) -> dict[str, list[str] | Any | int]:
...
+        Note that the samples will be instantiated with probabilities, so
+        we cannot come back after that to the StateVector amplitudes.
+
+        Returns:
+            A list of Samples representing the possible measurement outcomes 
+            of the state represented by this StateVector.
+
+        Example:
+            >>> state_vector = StateVector(np.array([1, 0, 0, -1])/np.sqrt(2), 2)
+            >>> state_vector.to_sample_list()
```
ThomasB-Colibri, line 124: Hi @joeffreyBelot, we apologise for the delay.

The doctest fails on CI because the computed probability is 0.4999999999999999 rather than exactly 0.5. Could you use # doctest: +ELLIPSIS for the sample-list representation ?

>>> state_vector.to_sample_list() # doctest: +ELLIPSIS
[Sample(2, index=0, count=None, probability=...), Sample(2, index=3, count=None, probability=...)]

PR #209, opened 2026-08-05
Feat quantinuum provider integration

MoHermes:
Adds Quantinuum backend support to MPQP using both Quantinuum Nexus and local TKET simulators.

### What’s included
- Added the optional Quantinuum dependencies, installable with ``pip install mpqp[quantinuum]``
- Added QUANTINUUMDevice with support for simulators and emulators
    - ``Local TKET Aer``, ``Aer State`` and ``Qulacs`` simulators
    - Quantinuum ``H-Series`` noiseless simulators and noisy emulators
    - Nexus-hosted ``Aer`` and ``Qulacs`` simulators
- Added Quantinuum Nexus account configuration through ``setup_connections``, then authentication using ``qnx.login()``
- Added MPQP Qcircuit conversion to TKET circuits for Quantinuum execution
- Added TKET translation for MPQP Pauli strings and observables using ``QubitPauliOperator``

#### Supported job types
- SAMPLE jobs
- STATE_VECTOR jobs
- OBSERVABLE jobs
    - Exact observables
    - Shot-based observables
    - Remote observable submission and retrieval


mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -283,20 +284,35 @@ def to_other_language(
...
     def to_other_language(
         self, language: Language, circuit: Optional[CirqCircuit] = None
     ) -> Union[
-        SparsePauliOp, QLMObservable, Hermitian, CirqPauliSum, CirqPauliString
+        SparsePauliOp,
+        QLMObservable,
+        Hermitian,
+        CirqPauliSum,
+        CirqPauliString,
+        QubitPauliOperator,
```
hJaffaliColibritd, line 310: Should we add `Sum` here to match the definition of the method? (line 312)

mpqp/execution/devices.py
```diff
@@ -546,3 +547,65 @@ def supports_observable(self) -> bool:
...
+    TKET_AER_SIMULATOR = "tket-aer"
+    TKET_AER_STATE_SIMULATOR = "tket-aer-state"
+    TKET_QULACS_SIMULATOR = "tket-qulacs"
+
+    NEXUS_AER_SIMULATOR = "aer"
+    NEXUS_AER_STATE_SIMULATOR = "aer-state"
+    NEXUS_QULACS_SIMULATOR = "qulacs"
+
+    H1_1LE = "H1-1LE"
+    H2_1LE = "H2-1LE"
```
hJaffaliColibritd, line 565: Since we can't test yet the execution on real hardware, should we put them in the current version of the code?
Or we let the user who has QPU access discover if it works or not?

mpqp/execution/remote_handler.py
```diff
@@ -12,25 +12,30 @@
...
 from mpqp.execution.providers.azure import get_result_from_azure_job_id
 from mpqp.execution.providers.ibm import get_result_from_ibm_job_id
+from mpqp.execution.providers.quantinuum import get_result_from_quantinuum_job_id
 
 
 def get_remote_result(
     job_data: str | Job, device: Optional[AvailableDevice] = None
 ) -> Result:
-    """Retrieve and parse a remote the result from a job_id and device. If the
+    """Retrieve and parse a remote result from a job ID and device. If the
```
hJaffaliColibritd: It retrieves the result from a job_id or a `Job`, we should maybe clarify it in the doc

hJaffaliColibritd (changes requested):
Partial review, will continue later

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,796 @@
...
+
+    if job.job_type == JobType.OBSERVABLE:
+        if TYPE_CHECKING:
+            assert isinstance(job.measure, ExpectationMeasure)
+        if job.measure.shots > 0:
+            return run_quantinuum_observable(job)
+
+    from pytket.extensions.qiskit.backends.aer import AerBackend, AerStateBackend
+    from pytket.extensions.qiskit.qiskit_convert import qiskit_to_tk
+    from pytket.extensions.qulacs.backends.qulacs_backend import QulacsBackend
```
hJaffaliColibritd: To be imported only when needed, if the user selects Aer, then no need to import qulacs, and vice versa

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,796 @@
...
+    from pytket.utils.operators import QubitPauliOperator
+    from qiskit import QuantumCircuit
+    from qnexus.models.references import (
+        CircuitRef,
+        CompilationResultRef,
+        ExecuteJobRef,
+        ExecutionResultRef,
+    )
+
+
```
hJaffaliColibritd, line 35: Would be nice to add the quantinuum devices and execution modes to the test suite we have already involving all the providers. Will allow to quickly check if all computations (at least locally) are correct and catch easy bugs.

mpqp/core/instruction/measurement/expectation_value.py
MathieuG-Colibri: done [6454def](https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/6454def4e427bb253fa3b34857f2fa0cad7d1b40)

mpqp/execution/devices.py
```diff
@@ -546,3 +548,65 @@ def supports_observable(self) -> bool:
...
+            QUANTINUUMDevice.NEXUS_AER_STATE_SIMULATOR,
+            QUANTINUUMDevice.TKET_AER_STATE_SIMULATOR,
+        }
+
+    def supports_state_vector(self) -> bool:
+        return self in {
+            QUANTINUUMDevice.NEXUS_AER_STATE_SIMULATOR,
+            QUANTINUUMDevice.NEXUS_QULACS_SIMULATOR,
+            QUANTINUUMDevice.TKET_AER_STATE_SIMULATOR,
+            QUANTINUUMDevice.TKET_QULACS_SIMULATOR,
```
hJaffaliColibritd: Can we double check this (that both don't support statevector)? I am surprized that QULACS simulator doesn't support statevector, it was in particular known to be the fastest in this domain. Also AER simulator in Qiskit supports state vector, but is just slower.
The fact that we can do ideal observable simulation also let me think we can do state vector with these devices.

mpqp/execution/devices.py
MoHermes: If I understand your concern correctly, ``H1-1LE`` and ``H2-1LE`` are not real hardware devices. They are noiseless emulators, and they can be tested without QPU access. The actual ``H1`` and ``H2`` hardware targets are not included in the current implementation

mpqp/execution/devices.py
MoHermes: Both support state-vector
``QulacsConfig(result_type="state_vector", ...)`` by default from the official quantinuum documentation
For Aer:
``TKET_AER_SIMULATOR`` uses the sampling oriented ``AerBackend``
``TKET_AER_STATE_SIMULATOR`` uses ``AerStateBackend`` and supports state vectors.
Qulacs support both capabilities, ideal observable and state vector, but Aer uses separate sampling and state vector backends, so supporting ideal observable evaluation does not automatically support retrieving the state vector result

mpqp/execution/providers/quantinuum.py
MoHermes: Done 1675615

mpqp/execution/remote_handler.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/4d764ccb170c0b6bbf02155e8895ac806d6f4e13

docs/getting-started.rst
```diff
@@ -260,6 +270,58 @@ script and select AWS configuration, you will have to choose between one of the
...
+<https://docs.quantinuum.com/nexus/trainings/notebooks/basics/auth_quotas.html>`_
+for more information.
+
+After installing the ``quantinuum`` extra, run the connection setup script:
+
+.. code-block:: console
+
+    $ setup_connections
+
+Select ``Quantinuum``. MPQP calls ``qnx.login()`` to start the Nexus
```
hJaffaliColibritd, line 287:
```suggestion
Select ``Quantinuum Nexus``. MPQP calls ``qnx.login()`` to start the Nexus
```

mpqp/execution/connection/quantinuum_connection.py
```diff
@@ -0,0 +1,207 @@
...
+
+    _activate_quantinuum_project()
+    return [str(job.id) for job in qnx.jobs.get_all()]
+
+
+def delete_quantinuum_account():
+    """Delete the locally stored Quantinuum Nexus configuration."""
+    decision = input(
+        colored(
+            "This will delete the local Quantinuum configuration and log out from Nexus. Continue? [y/N] ",
```
hJaffaliColibritd, line 186:
```suggestion
            "This will delete the local Quantinuum Nexus configuration and log out from Nexus. Continue? [y/N] ",
```

mpqp_scripts/setup_connections.py
```diff
@@ -122,6 +137,7 @@ def main_setup():
             AnswerNode("Amazon Braket", setup_aws_braket_account),
             AnswerNode("IonQ", config_ionq_key),
             AnswerNode("Azure", config_azure_account),
+            AnswerNode("Quantinuum", setup_quantinuum_account),
```
hJaffaliColibritd, line 140:
```suggestion
            AnswerNode("Quantinuum Nexus", setup_quantinuum_account),
```

mpqp_scripts/setup_connections.py
```diff
@@ -93,6 +104,7 @@ def delete_all():
             AnswerNode("Braket", delete_aws_braket_account),
             AnswerNode("IonQ", delete_ionq_account),
             AnswerNode("Azure", delete_azure_account),
+            AnswerNode("Quantinuum", delete_quantinuum_account),
```
hJaffaliColibritd, line 107:
```suggestion
            AnswerNode("Quantinuum Nexus", delete_quantinuum_account),
```

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
+from __future__ import annotations
```
hJaffaliColibritd, line 1: Didn't know where to put it
It execute the observale ideal job, but we have a runtime error concerning Context

Image

docs/getting-started.rst
```diff
@@ -260,6 +270,58 @@ script and select AWS configuration, you will have to choose between one of the
...
+project name in its local configuration and activates that project when a
+remote job is submitted. Running the setup again allows you to authenticate
+again or select another project. More information about Nexus projects is
+available in the `QNexus projects API reference
+<https://docs.quantinuum.com/nexus/nexus_api/projects.html>`_.
+
+MPQP provides local TKET simulators that do not require Nexus authentication.
+The Nexus connection is required for the remote Nexus-hosted Aer and Qulacs
+simulators, the noiseless H-Series ``QUANTINUUMDevice.H1_1LE`` and
+``QUANTINUUMDevice.H2_1LE`` emulators, and the noisy H-Series
```
hJaffaliColibritd:
```suggestion
``QUANTINUUMDevice.H2_1LE`` simulators, and the noisy H-Series
```

docs/getting-started.rst
```diff
@@ -260,6 +270,58 @@ script and select AWS configuration, you will have to choose between one of the
...
+for details about the corresponding Nexus backends.
+
+For example, the following circuit is executed remotely on the Nexus-hosted
+Aer simulator:
+
+.. code-block:: python
+
+    from mpqp import BasisMeasure, H, QCircuit, QUANTINUUMDevice, run
+
+    circuit = QCircuit([H(0), BasisMeasure([0], shots=100)])
```
hJaffaliColibritd:
```suggestion
    circuit = QCircuit([H(0), BasisMeasure(shots=100)])
```

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -280,23 +281,48 @@ def to_other_language(
...
+    @overload
     def to_other_language(
         self, language: Literal[Language.QASM2, Language.QASM3]
     ) -> Never: ...
     @overload
     def to_other_language(
-        self, language: Language, circuit: Optional[CirqCircuit] = None
+        self,
+        language: Language,
+        targets: Optional[list[int]] = [],
```
hJaffaliColibritd: Don't initialize here the default parameter. Put to `None` and then in the body `if targets is None then targets = []`

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -348,7 +374,9 @@ def to_other_language(
                     ),
                 )
         elif language == Language.CIRQ:
-            return self.pauli_string.to_other_language(Language.CIRQ, circuit)
+            return self.pauli_string.to_other_language(Language.CIRQ, circuit=circuit)
+        elif language == Language.TKET:
+            return self.pauli_string.to_other_language(Language.TKET, targets=targets)
```
hJaffaliColibritd:
```suggestion
            if targets is None:
                targets = []
            return self.pauli_string.to_other_language(Language.TKET, targets=targets)
```

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -731,9 +732,11 @@ def from_other_language(  # TODO: better typing using overloads
                     "Cannot parse non-homogeneous types when `pauli` is a `list`."
                 )
         from mpqp.environment.var_cache import (
-            InstalledProviders,
             _INSTALLED_MPQP_PROVIDERS,  # pyright: ignore[reportPrivateUsage]
         )
+        from mpqp.environment.var_cache import (
+            InstalledProviders,
+        )
```
hJaffaliColibritd: Why moving it to another import with same location ?

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -814,10 +829,14 @@ def to_other_language(
...
         list[Term],
+        QubitPauliOperator,
     ]: ...
 
     def to_other_language(
-        self, language: Language, circuit: Optional[CirqCircuit] = None
+        self,
+        language: Language,
+        targets: Optional[list[int]] = None,
+        circuit: Optional[CirqCircuit] = None,
```
hJaffaliColibritd, line 839: explain arguments in the doc, when do we use targets and circuit

mpqp/core/instruction/measurement/expectation_value.py
```diff
@@ -280,23 +281,48 @@ def to_other_language(
...
+        CirqPauliSum,
+        CirqPauliString,
+        QubitPauliOperator,
     ]: ...
 
     def to_other_language(
-        self, language: Language, circuit: Optional[CirqCircuit] = None
+        self,
+        language: Language,
+        targets: Optional[list[int]] = None,
```
hJaffaliColibritd, line 316: argument missing in the docstring

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -880,14 +900,51 @@ def to_other_language(
...
+            from pytket.circuit import Qubit
+            from pytket.pauli import Pauli, QubitPauliString
+            from pytket.utils.operators import QubitPauliOperator
+
+            pauli_gate_map = {
+                "I": Pauli.I,
+                "X": Pauli.X,
+                "Y": Pauli.Y,
+                "Z": Pauli.Z,
+            }
```
hJaffaliColibritd, line 924: Shouldn't this be handled at the PauliAtom level, then PauliMonomial level (to_other_language) ?
having a function that return the right pytket.pauli.Pauli for each PauliAtom
and then the right list of pytket.pauli.Pauli for each PauliMonomial, and then just calling that in the to_other_language of the Paulistring ?
This could simplify a bit the different level that we handle directly here ?

or maybe because of the specific targets it becomes less convenient? to be discussed

mpqp/core/instruction/measurement/pauli_string.py
```diff
@@ -1651,6 +1741,8 @@ def to_other_language(
             return pauli_gate_map[self.label](
                 LineQubit(0) if target is None else target
             )
+        elif language == Language.TKET:
+            return PauliString.to_other_language(self, Language.TKET, targets=targets)
```
hJaffaliColibritd, line 1747: For PauliAtom I would have returned a `pytket.pauli.Pauli` object no ?
This goes in the direction of separating what we do for a PauliString vs what we do for PauliAtom, and tink for PauliMonomial what is the better corresponding translation

mpqp/core/circuit.py
hJaffaliColibritd, line 1453: don't we have already a piece of code that does this somewhere in the runner or when we generate the job? I feel like recently we've been copy pasting this code everywhere

mpqp/execution/devices.py
hJaffaliColibritd: Indeed you right, I resolve

tests/core/instruction/measurement/test_expectation_value.py
```diff
@@ -49,3 +49,26 @@ def test_to_other_language_cirq(
 ):
     for obs, translation in list_to_cirq_pauli:
         assert obs.to_other_language(Language.CIRQ) == translation
+
+
+@pytest.mark.provider("quantinuum")
+def test_to_other_language_tket():
```
hJaffaliColibritd, line 55: Could be super nice to have a test with this story of `targets` to test that it is handled well when we don't have all qubits

tests/execution/connection/test_quantinuum_connection.py
```diff
@@ -0,0 +1,189 @@
...
+    assert result == ("Quantinuum account deleted.", [])
+    qnx_logout.assert_called_once_with()
+    assert quantinuum_config == {
+        "QUANTINUUM_CONFIGURED": "False",
+        "QUANTINUUM_PROJECT_NAME": "",
+    }
+
+
+@pytest.mark.skipif(
+    "--long" not in sys.argv,
```
hJaffaliColibritd, line 185: In principle we run those tests only in the remote case, so I would expect something like --long-costly or --long-remote,

tests/execution/providers/test_quantinuum.py
```diff
@@ -0,0 +1,251 @@
+"""Add ``--long`` to the CLI arguments to run the remote tests (disabled by
```
hJaffaliColibritd, line 1: In principle we run those tests remotely only under the --long-costly argument.
We need to update this whole file otherwise it could trigger remote runs if we just launch --long or --long-local tests, right?

requirements-dev.txt
```diff
@@ -1,9 +1,8 @@
--r requirements.txt
-setuptools==81.0.0
+-r requirements-all.txt
+setuptools
```
hJaffaliColibritd: better to fix the version, but maybe to a more recent one. see this PR : https://github.com/ColibrITD-SAS/mpqp/pull/214

mpqp/execution/runner.py
```diff
@@ -317,7 +319,18 @@ def _run_single(
...
+                ):
+                    from warnings import warn
+
+                    warn(
+                        f"The diagonal observable(s) cannot be optimized on the device: {device} due to the optimization process changing the jobtype to an incompatible one."
+                    )
+                else:
+                    return _run_diagonal_observables(
+                        circuit, measure, device, job, values
+                    )
```
hJaffaliColibritd, line 378: I'm not completely sure of that because in principle in _run_diagonal_observables, if the job is statevector we take the probabilities from the statevector and we compute the expectation value (which is a dot product of probas with diagonal coefs of the observable).
and if the job is not a statevector one, but sample, we sample the state and geenrate quasi-probabilities from the counts, and then do the same thing (dot product with obs coefs)
So in principle we don't change the job type, unless I missed something

tests/execution/connection/test_quantinuum_connection.py
```diff
@@ -0,0 +1,189 @@
+# pyright: reportPrivateUsage=false
```
hJaffaliColibritd, line 1: all these tests require that we have configured (or have credentials of) NExus account right ? so they should be launch under some specific pytest arguments ?

mpqp/execution/providers/quantinuum.py
hJaffaliColibritd: Done, I resolve

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    from qnexus.models.references import (
+        CircuitRef,
+        CompilationResultRef,
+        ExecuteJobRef,
+        ExecutionResultRef,
+    )
+
+
+def run_quantinuum(job: Job) -> Result:
+    """Execute the job on the selected Quantinuum device (local or remote),
```
hJaffaliColibritd:
```suggestion
    """Executes the job on the selected Quantinuum device (local or remote),
```

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+        backend_result = result_ref.download_result()
+        if TYPE_CHECKING:
+            assert isinstance(backend_result, BackendResult)
+
+        return extract_result(backend_result, job)
+    except Exception as error:
+        job.status = JobStatus.ERROR
+        job.status_message = str(error)
+        raise
```
hJaffaliColibritd: I don't like this structure, in principle we have run remote or run local, and then inside the run_local/remote we call the right method depending on the job type etc, submission and waiting for results for remote, etc
he it's a bit a mess

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+        if not isinstance(job.device, QUANTINUUMDevice):
+            raise DeviceJobIncompatibleError(
+                "`job` must correspond to a `QUANTINUUMDevice`, but corresponds "
+                f"to a {job.device} instead."
+            )
+
+        if not job.device.is_remote():
+            return run_tket_local(job)
+        if job.job_type == JobType.OBSERVABLE:
+            check_job_compatibility(job)
```
hJaffaliColibritd: the check of the compatibility should in principle be done for all type of jobs, local, remote, observable or not

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    Note:
+        This function is not meant to be used directly, please use
+        :func:`~mpqp.execution.runner.run` instead.
+    """
+    try:
+        if not isinstance(job.device, QUANTINUUMDevice):
+            raise DeviceJobIncompatibleError(
+                "`job` must correspond to a `QUANTINUUMDevice`, but corresponds "
+                f"to a {job.device} instead."
+            )
```
hJaffaliColibritd: This check is already done in check_job_compatibility, we need to call the function here

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+        )
+
+    if job.job_type == JobType.STATE_VECTOR and not job.device.supports_state_vector():
+        raise DeviceJobIncompatibleError(
+            f"{job.device} does not support `STATE_VECTOR` jobs."
+        )
+
+    if (
+        job.job_type == JobType.OBSERVABLE
+        and isinstance(job.measure, ExpectationMeasure)
```
hJaffaliColibritd: I think this check was already done before, if job type is OBSERVABLE then the job meaasure is necessarly of type ExpectationMeasure.

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+        and job.measure.shots == 0
+    ):
+        supports_exact_observable = (
+            job.device.supports_state_vector()
```
hJaffaliColibritd: In principle the method `supports_observable_ideal` should give you directly the results, no matter if the device is local or remote.

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+            raise DeviceJobIncompatibleError(
+                f"{job.device} requires a positive number of shots for "
+                "observable jobs."
+            )
+
+    if (
+        job.job_type == JobType.OBSERVABLE
+        and isinstance(job.measure, ExpectationMeasure)
+        and job.measure.shots > 0
+        and (not job.device.supports_observable() or not job.device.supports_samples())
```
hJaffaliColibritd: in principle `supports_observable` give you directly the info no ? otherwise why do we have a separate `supports_observable_ideal`?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+        )
+
+    if (
+        job.job_type == JobType.OBSERVABLE
+        and isinstance(job.measure, ExpectationMeasure)
+        and job.measure.shots > 0
+        and job.measure.commuting_type != CommutingTypes.QUBITWISE
+    ):
+        raise DeviceJobIncompatibleError(
+            "Quantinuum sampld observable jobs currently require qubit-wise "
```
hJaffaliColibritd:
```suggestion
            "Quantinuum sampling observable jobs currently require qubit-wise "
```

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+def run_tket_local(job: Job) -> Result:
+    """Execute a job using a local TKET backend.
+
+    Args:
+        job: Job targeting a local TKET device.
+
+    Returns:
+        The result after local compilation and execution.
+    """
+    check_job_compatibility(job)
```
hJaffaliColibritd, line 161: this should be done once in `run_quantinuum` and that's it

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    if TYPE_CHECKING:
+        assert isinstance(job.device, QUANTINUUMDevice)
+    if job.device.is_remote():
+        raise ValueError("The job must target a local TKET device.")
```
hJaffaliColibritd: This was already checked because we are in the run_tket_local

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+    if job.job_type == JobType.OBSERVABLE:
+        if TYPE_CHECKING:
+            assert isinstance(job.measure, ExpectationMeasure)
```
hJaffaliColibritd: This check whas already done previously in principle we can remove it

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+    if job.circuit.transpiled_circuit is None:
+        tket_circuit = job.circuit.to_other_device(job.device)
+    else:
+        from pytket.extensions.qiskit.qiskit_convert import qiskit_to_tk
+
+        qiskit_circuit = job.circuit.transpiled_circuit
+        if TYPE_CHECKING:
+            assert isinstance(qiskit_circuit, QuantumCircuit)
+        tket_circuit = qiskit_to_tk(qiskit_circuit)
```
hJaffaliColibritd: In one case we treat the transpiled circuit as a tket circuit, in the other as a qiskit circuit ? I would have save the transpiled_circuit as tket one to avoid retranslting from qiskit to tket each time.

But it seems here that we didn't merge the feature that handles transpiled circuit for different providers (as a dictionnary and not a single variable).

But most importantly it seems that we never set actually the transpiled_circuit of the circuit, neither in to_other_language nor to_other_device for Quantinuum ....

I would either handle that correctly here, or just not do it and put a TODO to do it properly  later.

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    elif job.device == QUANTINUUMDevice.TKET_QULACS_SIMULATOR:
+        from pytket.extensions.qulacs.backends.qulacs_backend import QulacsBackend
+
+        backend = QulacsBackend()
+    else:
+        raise ValueError(f"Local TKET device {job.device} is not handled.")
+
+    compiled_circuit = backend.get_compiled_circuit(
+        tket_circuit,
+        optimisation_level=0,
```
hJaffaliColibritd: In the future we should allow the user to tune the optimisation level, especially for real hardware run.

This can be done already today because we merged in dev the "Provider params feature"

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    )
+    if job.job_type == JobType.OBSERVABLE:
+        return run_tket_observable(job, compiled_circuit, backend)
+
+    if job.job_type == JobType.SAMPLE:
+        if TYPE_CHECKING:
+            assert isinstance(job.measure, BasisMeasure)
```
hJaffaliColibritd: check already done in principle, not necessary

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+        n_shots = job.measure.shots
+    else:
+        n_shots = None
```
hJaffaliColibritd:
```suggestion
    n_shots = job.measure.shots if job.job_type == JobType.SAMPLE else None
```

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    backend_result = backend.run_circuit(compiled_circuit, n_shots=n_shots)
+    return extract_result(backend_result, job)
+
+
+def run_tket_observable(
+    job: Job,
+    tket_circuit: "TKETCircuit",
+    backend: "Backend",
+) -> Result:
+    """Return the result of an exact `OBSERVABLE` job using the built in
```
hJaffaliColibritd: I have a problem that we handle here only the "ideal" expectaiton value, while we can in principle arrive here if we have Observable job with shot noise ....
It seems that this case is completely discarded here ? In the same time, there is no distinction of cases in `run_tket_local` depending on the value of the shot to redirect to another method for observable sampled job.
I am confused

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    Args:
+        job: Job to execute.
+        tket_circuit: Compiled TKET circuit on which the observables are evaluated.
+        backend: Local TKET backend used to evaluate the observables.
+
+    Returns:
+        A result containing the exact expectation values of the observables.
+    """
+    if not isinstance(job.measure, ExpectationMeasure):
+        raise ValueError("Observable jobs must have an `ExpectationMeasure`.")
```
hJaffaliColibritd: Again and again we check ?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+        operator = observable.to_other_language(Language.TKET, job.measure.targets)
+        if TYPE_CHECKING:
+            assert isinstance(operator, QubitPauliOperator)
+        expectation_value = backend.get_operator_expectation_value(
+            tket_circuit,
+            operator,
+        )
+        expectation_values[label] = float(expectation_value.real)
+
+    return extract_observable_result(job, expectation_values, 0.0, 0)
```
hJaffaliColibritd: This tells me that error and shots are equal to 0, so this function only handle the observable ideal case, and should be renamed accordingly
Where do we handle the local observable case with shot noise ...?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+
+def run_quantinuum_observable(job: Job) -> Result:
```
hJaffaliColibritd: I am completely confused about the organization of the running functions between local or remote, ideal or not ideal, run functions or different categories calling each others ... I think we need to reorganize this.
Here it's not clear if we are doing remote or local observable jobs.

if we mean remote by "quantinuum" in the name, then I will replace by nexus to be more clear, because everything here is quantinuum in principle.
If we mean observable in general, then why this method is not direclty called when we detect observable job, and why this method do not call run_tket_observable ...

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    """Execute an observable job using a supported Quantinuum backend.
+
+    Exact observables are computed from a state vector when `shots=0`.
```
hJaffaliColibritd: I am a bit confused here. Why do we call a state vector to compute exact observables ... Either the backend supports ideal observable or not. And if not in principle it doesn't support statevector right ?

If we mix local cases here ... then we just call run_tket_observable for ideal no ?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    observable_labels = job.measure.observables_labels
+
+    circuit = job.circuit.without_measurements()
+
+    if job.measure.shots == 0:
+        state_job = Job(
+            JobType.STATE_VECTOR,
+            circuit,
+            job.device,
+        )
```
hJaffaliColibritd: I'm not fan of this ... changing type of job on the go without warning the user. And do we really need it ? If backend supports statevector, in principle it handles ideal observable no ? Or we have exceptions ? 

In the second case, do we really want to manually fix that on our side ?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+        expectation_values: dict[str, float] = {}
+        for label, observable in zip(
+            observable_labels,
+            job.measure.observables,
+        ):
+            expectation_values[label] = float(
+                np.vdot(
+                    state_result.amplitudes,
+                    observable.matrix @ state_result.amplitudes,
+                ).real
```
hJaffaliColibritd: I can't believe we have to compute the expectation value by hand using the big matrices, and computed locally .... this is not viable, especially if we are talking about remote executions here .... it's not to do the job half remote and half locally....

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+        )
+        sample_job = Job(JobType.SAMPLE, sample_circuit, job.device)
+        sample_result = run_quantinuum(sample_job)
+        job.id = sample_job.id
+        grouped_probabilities.append(sample_result.probabilities)
+
+    return extract_sampled_observable_result(job, grouped_probabilities)
+
+
+def submit_job_quantinuum(job: Job) -> tuple[str, "ExecuteJobRef"]:
```
hJaffaliColibritd: Rename submit_job_nexus ?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+    return submit_circuits_to_nexus(
+        job,
+        [job.circuit],
+        n_shots,
+        name=f"mpqp-{job.job_type.name.lower()}-{job.device.value}",
+    )
+
+
+def submit_quantinuum_observable(job: Job) -> tuple[str, "ExecuteJobRef"]:
```
hJaffaliColibritd: Rename submit_nexus_observable

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    each qubit-wise commuting Pauli group is submitted as a circuit containing
+    the required basis change followed by a measurement. The original
+    MPQP `Job` is required to reconstruct the expectation value from the
+    returned states or counts.
+    """
+    check_job_compatibility(job)
+    if TYPE_CHECKING:
+        assert isinstance(job.device, QUANTINUUMDevice)
+        assert isinstance(job.measure, ExpectationMeasure)
+    if not job.device.is_remote():
```
hJaffaliColibritd: in principle this is already checked in the function `submit` from runner.py, which is the one calling the remote submission to quantinuum nexus, that is calling this function.

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+        raise ValueError("Observable submission requires a remote Quantinuum device.")
+
+    circuit = job.circuit.without_measurements()
+    n_shots: int | list[None]
+    if job.measure.shots == 0:
+        circuits = [circuit]
+        n_shots = [None]
```
hJaffaliColibritd: Why do we need to send a statevector job remotely for an Ideal Observable job?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    else:
+        from mpqp.tools.pauli_grouping import find_qubitwise_rotations
+
+        circuits = []
+        for group in job.measure.get_pauli_grouping():
```
hJaffaliColibritd: better to save the groups generated by the pauli grouping to reconstruct later the results no ?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+    circuit = job.circuit.without_measurements()
+    n_shots: int | list[None]
+    if job.measure.shots == 0:
+        circuits = [circuit]
+        n_shots = [None]
+    else:
+        from mpqp.tools.pauli_grouping import find_qubitwise_rotations
+
+        circuits = []
```
hJaffaliColibritd, line 509: In principle we don't do systematically Pauli grouping and optimization of measurements.
It will depend on the boolean attribute `expectation_measure.optimize_measurement`. So i would handle also the case when we don't do pauli grouping.
I would also put a warning to recall that we will send different type of jobs when we do pauli grouping.

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+            circuits.append(sample_circuit)
+        n_shots = job.measure.shots
+
+    return submit_circuits_to_nexus(
+        job,
+        circuits,
+        n_shots,
+        name=f"mpqp-observable-{job.device.value}",
+        description="mpqp:observable",
+    )
```
hJaffaliColibritd: this means that if the user submits a job of observable he will get job ids and execution refs of circuits he doesn't really know to what they correspond ...
Maybe we should document better what is happening, or associate a circuit rref with the correspond group in the return type ? 
This requires better thinking from the user point of view to be able to retrive the observable job result then from pauli grouping sampling circuits

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+
+def submit_circuits_to_nexus(
+    job: Job,
+    circuits: list[QCircuit],
+    n_shots: int | list[None],
+    name: str,
+    description: str = "",
+) -> tuple[str, "ExecuteJobRef"]:
+    """Upload, compile, and submit several circuits as one Nexus execute job."""
```
hJaffaliColibritd: Better docstring

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+    if not isinstance(job.device, QUANTINUUMDevice) or not job.device.is_remote():
+        raise ValueError("Nexus submission requires a remote Quantinuum device.")
```
hJaffaliColibritd: This is not necessary in principle, sinc the user is not supposed to call directly this function, and we normally come from `mpqp.execution.runner.submit`

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+    tket_circuits = []
+    for circuit in circuits:
+        if circuit.transpiled_circuit is None:
+            tket_circuits.append(circuit.to_other_device(job.device))
+        else:
+            qiskit_circuit = circuit.transpiled_circuit
+            if TYPE_CHECKING:
+                assert isinstance(qiskit_circuit, QuantumCircuit)
+            tket_circuits.append(qiskit_to_tk(qiskit_circuit))
```
hJaffaliColibritd: Again I don't like this, either we save quantinuum circuit already translated, or we precise that we check both quantinuum circuit transpiled and qiskit circuit also, but for that you need the dictionnary logic.
Also because one translated/transpiled; we don't save it in the attribute so we will apparently repeat the translation each time. not fan

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+            circuit=tket_circuit,
+            name=f"{name}-circuit-{index}",
+        )
+        for index, tket_circuit in enumerate(tket_circuits)
+    ]
+
+    compile_job_ref = qnx.start_compile_job(
+        programs=uploaded_circuit_refs,
+        backend_config=backend_config,
+        optimisation_level=0,
```
hJaffaliColibritd: Allow user to configure the optimization level, or add a TODO/ Notion card)

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+    job.status = JobStatus.RUNNING
+
+    execute_job_ref = qnx.start_execute_job(
+        programs=compiled_circuit_refs,
+        backend_config=backend_config,
+        n_shots=n_shots,
+        name=f"{name}-execution-job",
+        description=description,
+    )
```
hJaffaliColibritd, line 639: This indicate we can do multi-circuit / batch execution @MathieuG-Colibri @JulienCalistoTD on Quantinuum Nexus with a single job ID ?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    return job.id, execute_job_ref
+
+
+def extract_observable_result(
+    job: Job,
+    expectation_values: dict[str, float],
+    errors: float | dict[str, float],
+    shots: int,
+) -> Result:
+    """Construct an MPQP result from Quantinuum expectation values."""
```
hJaffaliColibritd: better doctring

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+
+def extract_sampled_observable_result(
+    job: Job,
+    grouped_probabilities: list[npt.NDArray[np.float64]],
+) -> Result:
+    """Reconstruct a sampled observable result from one probability vector per
+    Pauli group by combining the measured probabilities with the eigenvalues and
+    coefficients of the observable's Pauli terms.
+    """
```
hJaffaliColibritd: better doctring

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    if not isinstance(job.measure, ExpectationMeasure):
+        raise ValueError("Observable jobs must have an `ExpectationMeasure`.")
```
hJaffaliColibritd: Not needed, this function is called already after checking everything and running

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+
+    grouping = job.measure.get_pauli_grouping()
```
hJaffaliColibritd: To avoid recomputing the pauli grouping I would pass the one already computed before submitting the job, or to create an attribute in the class ExpectationMeasure to store it if one needs to reuse it

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,800 @@
...
+    if len(grouped_probabilities) != len(grouping):
+        raise ValueError(
+            "The number of Quantinuum observable results "
+            f"({len(grouped_probabilities)}) does not match the number of "
+            f"submitted Pauli groups ({len(grouping)})."
+        )
```
hJaffaliColibritd: I think this should be checked right after the execution and retrival of quasi-probabilities from the counts, not here.

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,795 @@
...
+
+    from mpqp.tools.pauli_grouping import pauli_monomial_eigenvalues
+
+    eigenvalues = [
+        {monomial.name: pauli_monomial_eigenvalues(monomial) for monomial in group}
+        for group in grouping
+    ]
```
hJaffaliColibritd: normally each monomial is only present in one group, so we can have only one dictionnary, instead of a list of dictionnaries

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,795 @@
...
+
+    expectation_values: dict[str, float] = {}
+    errors: dict[str, float] = {}
+    for label, observable in zip(
```
hJaffaliColibritd: i think we can optimize this code, it's sufficient to take the monomials in each observable and go grab the corresponding eigen values in the dictionnary set earlier (which is currently a list when I review this code), the right probabilities, and then construct the expectation value.
No need to go through each group, each monomial, etc

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,795 @@
...
+    return extract_observable_result(
+        job,
+        expectation_values,
+        errors,
+        job.measure.shots,
+    )
+
+
+def extract_state_vector_result(
+    amplitudes: "list[Complex] | npt.NDArray[np.complex128]",
```
hJaffaliColibritd: Why the tiping is in string ?

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,795 @@
...
+    if job is not None and job.job_type == JobType.OBSERVABLE:
+        if not isinstance(job.measure, ExpectationMeasure):
+            raise ValueError("Observable jobs must have an `ExpectationMeasure`.")
+        job.id = job_id
+        if job.measure.shots == 0:
+            if len(result_refs) != 1:
+                raise ValueError(
+                    "An exact Quantinuum observable job must return one state-vector "
+                    f"result, but returned {len(result_refs)}."
+                )
```
hJaffaliColibritd: Not fan of this, see comments above, it should return an expectation value, otherwise it doesn't support ideal observable,
 we are not going to do manually and locally matrix products of 20 or 30 qubits

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,795 @@
...
+            result_ref = result_refs[0]
+            if TYPE_CHECKING:
+                assert isinstance(result_ref, ExecutionResultRef)
+            backend_result = result_ref.download_result()
+            if TYPE_CHECKING:
+                assert isinstance(backend_result, BackendResult)
+            state = backend_result.get_state()
+            expectation_values = {
+                label: float(np.vdot(state, observable.matrix @ state).real)
```
hJaffaliColibritd: /!\

mpqp/execution/providers/quantinuum.py
```diff
@@ -0,0 +1,795 @@
...
+                    f"Quantinuum observable job '{job_id}' returned no sample counts."
+                )
+            probabilities = np.zeros(2**job.circuit.nb_qubits, dtype=np.float64)
+            for outcome, count in raw_counts.items():
+                index = int("".join(str(bit) for bit in outcome), 2)
+                probabilities[index] = count / shots
+            grouped_probabilities.append(probabilities)
+        return extract_sampled_observable_result(job, grouped_probabilities)
+
+    if job is None and job_ref.annotations.description == "mpqp:observable":
```
hJaffaliColibritd: How do we do if the user wants to retrive result from jobs he launched without using MPQP ? he will not have the annotations rigth ?

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Isn't what we do anyway with pauli grouping ? We get the counts online and do the product to get the exp_value.

Also this code is only for local jobs in case of ideal observable.

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Same issue when we do pauli grouping technically we go from an Observable job to a Sample one (or StateVector when ideal). I'm not sure how to handle the user in this case, should they be told that their job is changed although we output the desired data ?

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Refactored run_observable to differentiate ideal from non ideal.
Also changed the function to avoir going back up the stack of the quantinuum runner.
https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/077cde374ff79f695d55752d6b4d95ad35bbd209

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: refactored to be more optimized and clear https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/077cde374ff79f695d55752d6b4d95ad35bbd209

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/077cde374ff79f695d55752d6b4d95ad35bbd209

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/077cde374ff79f695d55752d6b4d95ad35bbd209

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Done [077cde3](https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/077cde374ff79f695d55752d6b4d95ad35bbd209)

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: This part of the code was replaced.

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Yes we can, actually Muhammad uses this for sending the whole Pauli Grouping case at once which is pretty cool !

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: So this is done because in case we're running multiple time the same circuit we pre_transpile some information.
If we do a single dictionnary, we'll have to differentiate between groups of monoms.

Which we currently don't keep in pre_transpilation (for now) but we would need to do that to avoid calling get_pauli_grouping everytime.
So I suggest leaving it at that and putting a TODO because that doesn't really matter right now.

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Removed https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Maybe keeping that for another PR (I don't mind doing it here but it touches a lot different parts of the code)

Having a priority of if we already have Qiskit's version of the circuit the transpilation will get quicker however it's a bit touchy to do imo.

I'll let it be "if transpiled_circuit then t_circuit else translate it" for now.
https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Removed https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: The index of the circuit is the same as the index of the group in measurement.get_pauli_grouping this behavior has been documented as well and put in the description of the individual circuits.

https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Put an error if we try to go through nexus without optimize_measurement.

https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Done (also useful for other things imo) https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: It seems that we always do Pauli grouping here so for an ideal observable it's akin to a state_vector with extra steps.

Also I might be wrong but with tket it might be possible to compute a statevector as a remote job (purely classical job of course) but iirc they have big scale statevector computing which would make this case have some sense.

After checking in more detail: Pure Observable jobs remotely are not supported by Nexus our only option is to compute it ourselves.

Added saferails for these executions: https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Removed https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: still removed in the end https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Done https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/b6700abe128a8867fe1884403473812c027f86d3

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Yes but they can still provide the MPQP equivalent of the job (the same observables and targets).
It's not very good but I can't think of anything else https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/3571952d5c251556696da5f2fb34d53f4dbb41d6

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Changed it to the same computation we do with P grouping https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/3571952d5c251556696da5f2fb34d53f4dbb41d6

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: It doesn't really support observables...

However we can compute it using the PauliString which is way more manageable at high qubit counts.

That means that it's the exact opposite of what this ValueError is refering to, when we have observables we need a circuit for each groups period.

https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/3571952d5c251556696da5f2fb34d53f4dbb41d6

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: Idk removed https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/3571952d5c251556696da5f2fb34d53f4dbb41d6

mpqp/execution/providers/quantinuum.py
MathieuG-Colibri: The check has been moved (and changed at that) https://github.com/ColibrITD-SAS/mpqp/pull/209/commits/3571952d5c251556696da5f2fb34d53f4dbb41d6

mpqp/execution/devices.py
hJaffaliColibritd: @MathieuG-Colibri I suppose we double checked this in the recent implementations?

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Done

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Added task in Notion

mpqp/core/instruction/measurement/pauli_string.py
hJaffaliColibritd: Added task in notion

mpqp/core/circuit.py
hJaffaliColibritd: Added task in Notion

mpqp/execution/providers/quantinuum.py
hJaffaliColibritd: At the end ideal observable are not supported by MPQP directly. Because computing expectation value for each pauli monomial and each state, it can be comparable to matrix vector product. We can see later to activate this feature.

mpqp/execution/runner.py
MoHermes: The check is needed because the optimization requires either a state vector or sample execution and not every device supporting observable jobs supports those executions. I aslo updated the warning to make it more clear
7319033

tests/execution/connection/test_quantinuum_connection.py
MoHermes: Only ``test_remote_quantinuum_connection()`` needs real credentials, so ``--long`` is enough, it does not consume credits , so no need for ``--long-costly``, the other tests use mocked QNexus calls

mpqp/execution/providers/quantinuum.py
MoHermes: Can you run the example again?
On the current version, that example should stop with ``DeviceJobIncompatibleError`` before any Nexus job is submitted
image

PR #212, opened 2026-08-18
Perf separate measurements

MathieuG-Colibri:
Change QCircuit measurement methods to improve performances

### DIFF
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -304,8 +304,8 @@ class QCircuit:
                 if len(self.measurements) != 0:
                     raise InstructionAfterMeasurementError(
                         "Cannot add gate after measurement in the circuit."
-                        )
-                  
+                    )
+
             if isinstance(components, Measure):
                 self.measurements.append(components)
             else:
@@ -867,10 +867,7 @@ class QCircuit:
 
         """
         dagger = self._clone_without(
-            [
-                "instructions",
-                "measurements"
-            ],
+            ["instructions", "measurements"],
             deep_copy=True,
         )
 
@@ -1884,7 +1881,9 @@ class QCircuit:
 
     def __repr__(self) -> str:
         args = []
-        components: list[Instruction | NoiseModel] = self.instructions + self.measurements + self.noises
+        components: list[Instruction | NoiseModel] = (
+            self.instructions + self.measurements + self.noises
+        )
         if len(components) != 0:
             args.append(f"[{', '.join(repr(component) for component in components)}]")
         if self._user_nb_qubits is not None:
--- a/mpqp/core/circuit.py
+++ b/mpqp/core/circuit.py
@@ -55,7 +55,6 @@ from mpqp.core.languages import Language
 from mpqp.noise.noise_model import DimensionalNoiseModel, NoiseModel
 from mpqp.tools.errors import (
     DeviceJobIncompatibleError,
-    InstructionAfterMeasurementError,
     InstructionParsingError,
     NonReversibleWarning,
     NumberQubitsError,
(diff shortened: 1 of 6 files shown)


mpqp/core/circuit.py
```diff
@@ -160,6 +160,8 @@ def __init__(
         """See parameter description."""
         self.instructions: list[Instruction] = []
```
hJaffaliColibritd, line 160: Then if we remove the measurements from the instructions, this list is no more a list of instructions ?
Or maybe we can simplify it to Gate | BreakPoint | Barrier, whatever ? To be discussed

mpqp/core/circuit.py
```diff
@@ -299,14 +301,15 @@ def add(self, components: OneOrMany[Instruction | NoiseModel]):
...
-                            "Cannot add gate after measurement in the circuit."
-                        )
-                    if isinstance(self.instructions[i], Gate):
-                        break
-            self.instructions.append(components)
+                if len(self.measurements) != 0:
+                    raise InstructionAfterMeasurementError(
+                        "Cannot add gate after measurement in the circuit."
+                    )
+
```
hJaffaliColibritd: Can't we just remove this now ? Since we treat measurements separately, technically we suppose they always come at the end ? So people can add then whenver they want ? To be discussed, but it will change the current logic

mpqp/core/circuit.py
```diff
@@ -456,7 +459,7 @@ def _set_nb_qubits_dynamic(self, nb_qubits: int):
                 if noise._dynamic:  # pyright: ignore[reportPrivateUsage]
                     self._update_targets_components(noise)
 
-            for instruction in self.instructions:
+            for instruction in self.instructions + self.measurements:
```
hJaffaliColibritd: Maybe have a method to do that ? Something like self.instructions_with_measurements, I don't know

mpqp/core/circuit.py
```diff
@@ -754,7 +755,7 @@ def __len__(self) -> int:
             4
 
         """
-        return len(self.instructions)
+        return len(self.instructions + self.measurements)
```
hJaffaliColibritd: maybe faster to return `len(self.instructions) + len(self.measurements)` it avoids instantiating a new list with concatenation

hJaffaliColibritd (changes requested):

mpqp/core/circuit.py
MathieuG-Colibri: It's only used once and wouldn't really impact performances. So I'm down to keep it.

mpqp/core/circuit.py
MathieuG-Colibri: Done and yes your method has better performances.
[566a401](https://github.com/ColibrITD-SAS/mpqp/pull/212/commits/566a4016034623524d6fc21dd5c678afbca88a29)

mpqp/core/circuit.py
MathieuG-Colibri: removed [566a401](https://github.com/ColibrITD-SAS/mpqp/pull/212/commits/566a4016034623524d6fc21dd5c678afbca88a29)

mpqp/core/circuit.py
hJaffaliColibritd: I see it used several times (seven to be precise), and it seems to be something the user could ask for?
As you want. You can resolve this comment if we don't do it.

hJaffaliColibritd (approved):

mpqp/core/circuit.py
MathieuG-Colibri: done https://github.com/ColibrITD-SAS/mpqp/pull/212/commits/d00726dc68c2c13840cc5fb43d7cd883dab025a1

